PR #15, opened 2023-09-28
feat: CRM Task

shariquerik:
image

image

image


federicocalvo:
@shariquerik that its great! i am big fan of Daylite,  and a major differentiator is the tasks that helped to avoid skipping any stage of the sales process.
Captura de pantalla 2023-09-28 a la(s) 13 08 48

shariquerik:
> a major differentiator is the tasks that helped to avoid skipping any stage of the sales process.

@federicocalvo So according to your usecase you need to show task for each stages right?
Currently if you can see in task tab I am just showing all tasks
I am planning to add different views to it like
1. All tasks
2. Based on task status
Similarly we can add another view
3. Based on lead status

federicocalvo:
@shariquerik I was only referring to the fact that it is a great addition to have tasks by initiative, it helps a lot to predefine processes and not to forget certain issues.

Of course, it could be useful to have some differentiation by stage, or the different forms that you propose in the future.

Excellent work

PR #29, opened 2023-11-21
feat: Custom fields

shariquerik:
Show custom fields in lead/deal page

Added custom field **Stage** after **Industry**
image

Rendered custom field **Stage** on Lead Page
image

NOTE:
1. If you need to add a custom field add it using the customize form it will reflect in the CRM Lead/Deal page
2. To create more sections E.g. **Details** & **Person** add tabs in the doctype
3. If you do not want to show any tab as a section mark that tab `read_only` as you can see in the above ss there is **Others** tab in **CRM Lead** doctype but it is not visible on **Lead page** (fields inside the tab will also not be shown on the CRM page)


maveezmh:
if we add a dependable link field with filters. the filters are not working as expected

shariquerik:
> if we add a dependable link field with filters. the filters are not working as expected

Create a separate Issue with example and screenshots(or screen recording)

PR #185, opened 2024-05-17
feat: Clear all filter x button

shariquerik:
image


ajiragroup:
Is it possible to make phone number searchable ?

PR #214, opened 2024-06-07
feat: Add custom fields in quick entry modal

shariquerik:
image
image


shariquerik:
Currently it does not support all types of fields I will work on it

PR #293, opened 2024-08-09
feat: Preserve last visited tab

### LINKED ISSUE #192: Last Visited Tab
When i am on Whatsapp or any other tab and I refresh the browser tab, after refreshing it go to Activity tab

If you can set last visited tab for default tab for that lead and don't change tab when refresh,

gopikrishnan13:
- Implemented functionality to store and retrieve the last visited tab index in localStorage.
- The application now defaults to the last visited tab after a page refresh, rather than always returning to the Activity tab.
- This ensures that users remain on the same tab they were on before refreshing the page, improving user experience.

[Screencast from 09-08-24 12:45:17 PM IST.WEBM](https://github.com/user-attachments/assets/412d73fa-1660-4994-99ff-dbd8e1acf966)

Closes #192

### DIFF
--- a/frontend/src/composables/useActiveTabManager.js
+++ b/frontend/src/composables/useActiveTabManager.js
@@ -1,0 +1,68 @@
+import { ref, watch } from 'vue';
+
+// Debounce function to delay updates
+function debounce(fn, delay) {
+  let timeout;
+  return (...args) => {
+    clearTimeout(timeout);
+    timeout = setTimeout(() => fn(...args), delay);
+  };
+}
+
+export function useActiveTabManager(tabs, storageKey) {
+    
+  const preserveLastVisitedTab = debounce((tabName) => {
+    localStorage.setItem(storageKey, tabName.toLowerCase());
+  }, 300);
+
+  function setActiveTabInUrl(tabName) {
+    window.location.hash = '#' + tabName.toLowerCase();
+  }
+
+  function getActiveTabFromUrl() {
+    return window.location.hash.replace('#', '');
+  }
+
+  function findTabIndex(tabName){
+    return tabs.value.findIndex(tabOptions => tabOptions.name.toLowerCase() === tabName);
+  }
+
+  function getTabIndex(tabName) {
+    let index = findTabIndex(tabName)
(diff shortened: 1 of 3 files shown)


shariquerik:
@gopikrishnan13 This is not a correct way to fix this.
We should be able to change tabs via url
E.g `/crm/leads/CRM-LEAD-2024-0001#activity_tab`

And we can then store the last visited tab in local storage
So if url does not have `#activity_tab` in it then it will show last visited tab and if it has it then that tab will be loaded

gopikrishnan13:
Hi @shariquerik 

I made some updates based on your feedback and implemented the feature in the suggested way. it works fine.

I wrote this code in Lead.vue
image

Here’s the output video for your review.

[Screencast from 20-09-24 02:27:01 PM IST.webm](https://github.com/user-attachments/assets/8b9c9aa3-c193-4344-9895-29b21b55cc25)

Can I continue with this and make a commit?

Let me know if you have any further suggestions!

shariquerik:
@gopikrishnan13 Yeah go ahead

gopikrishnan13:
@shariquerik I updated this PR, please check it now,

- Finalized the implementation of useActiveTabManager for tab management.
- Enhanced logic to retrieve and set the active tab from both the URL and local storage.
- Set default behavior to the first tab if no valid active tab is found.
- Streamlined code for better readability and maintainability.

[Screencast from 25-09-24 06:00:23 PM IST.webm](https://github.com/user-attachments/assets/30053141-437a-41be-83a7-6b4d1310fd30)

shariquerik (approved):

shariquerik:
@gopikrishnan13 There is one problem occurred after this.
When you are in Deal Page and when you go back using browsers back button it does not go back it just removes #activity hash from the url and then again when you go back it goes back
And when you go forward it goes to /deals/<dealID>#activity and now you cannot go back it just lost history somehow

gopikrishnan13:
@shariquerik  I have another solution, I’ve tested it, and it works well. Instead of setting the hash directly, I replace the history state only the first time when setting the tab from local storage. This way, it doesn’t affect the browser history  and  this prevents adding extra history entries

Do you think this is a better solution?

image

image

Issue #284, opened 2024-08-02
Set proper default column


michellealva:
<img width="1437" alt="image" src="https://github.com/user-attachments/assets/04c9056c-0d53-49aa-b328-7dff88da7891">

Net to always reset it, set it by default.

PR #305, opened 2024-08-14
fix: default list columns doesn't work

shariquerik:
fixes: https://github.com/frappe/crm/issues/284


joshuarestivo:
Wow, that was quick! Thank you for addressing that.

PR #404, opened 2024-10-09
fix: Link field filters

saeedkola:
Custom fields with link filters defined were not reflecting on the CRM portal

the root cause is that the get_fields_layout API is not responding with the filters.
With this small change, it is working as expected


maveezmh:
@shariquerik

shariquerik (approved):

PR #410, opened 2024-10-12
fix: Restore defaults

shariquerik:
image


1ubuntuuser:
Thanks for adding the fix. I can't seem to find FCRM Settings anywhere. Where should that be?

1ubuntuuser:
After switching to the develop branch I found it by navigating to /app/frappe-crm

PR #427, opened 2024-10-24
feat: add support for incoming messages using quick reply buttons

saby22:
@shariquerik 

`frappe_whatsapp` recently added support for accepting incoming messages using quick reply buttons in [this](https://github.com/shridarpatil/frappe_whatsapp/pull/102) PR.

This just extends that feature to the CRM.


kumud-oa:
Looking forward to merging this. @shariquerik Please help with your review.

shariquerik (approved):

PR #443, opened 2024-11-11
Resolve performance issue with multiple assignees in MultipleAvatar component

Rajeshbala2020:


shariquerik:
@Rajeshbala2020 How it will solve performance?

PR #526, opened 2025-01-10
Enhancement - Feature to include all columns in listview

RenganathanRiver:
For user convenience, users can include all columns in one click


shariquerik:
@RenganathanRiver Does not make sense to add this, listview is used to see relevant fields only so giving access to add all the fields at once is misleading and will encourage users to make bad listviews.

PR #536, opened 2025-01-21
fix: stable tag for docker image

NagariaHussain:
Easy install commands do not work as per README.


revant (approved):

revant:
LMS diff:
https://github.com/frappe/lms/blob/e591cd74ab5583384c9c832a0f23124fd0e38475/.github/workflows/build.yml#L18-L19

PR #585, opened 2025-02-06
fix:  Cannot assign to "showSmallCallPopup" because it is a constant

kakde-sandeep:


kakde-sandeep:
Fixing the below error
Screenshot 2025-02-06 at 9 20 36 PM

shariquerik (approved):

Issue #594, opened 2025-02-18
Notifications trigger unnecessary API calls for all users


pu-raihan:
Currently, each time a CRM Notification is created or updated, a WebSocket event (crm_notification) is broadcasted to all users. As a result, the frontend reloads notifications for every user, even if the notification was intended for a single user.

###  **Steps to Reproduce:**
Create a new CRM Notification for a specific user.
Observe that all logged in users receive the event via WebSocket.
The `notifications.reload()` API call is triggered for everyone. This is not efficient for large scale production.
###  **Expected Behavior:**
Only the intended recipient should receive the WebSocket event.
Other users should not trigger unnecessary API calls.
###  **Proposed Fix:**
Modify `frappe.publish_realtime()` in crm_notification.py to send the event only to the recipient instead of broadcasting it globally.
Update the frontend WebSocket listener to ensure that only the logged-in user processes the event.
###  **Relevant Files:**
[crm/fcrm/doctype/crm_notification/crm_notification.py](https://github.com/frappe/crm/blob/main/crm/fcrm/doctype/crm_notification/crm_notification.py)
[frontend/src/components/Notifications.vue](https://github.com/frappe/crm/blob/main/frontend/src/components/Notifications.vue)

PR #596, opened 2025-02-19
Fix unnecessary notification API reloads for all users

pu-raihan:
This PR fixes the issue #594  where every user received a WebSocket event for all CRM notifications, causing unnecessary API reloads. Now, the event is emitted only for the intended recipient, reducing server load and improving efficiency.

###  **Changes Made:**
Updated `frappe.publish_realtime("crm_notification")` to send the event only to the relevant user.
<br>

###  **Before (Inefficient, Broadcasts to All Users)**
```py
frappe.publish_realtime("crm_notification")
 ```
<br>

###  **After (Optimized, Sends Only to Intended User)**
```py
frappe.publish_realtime("crm_notification", user=self.recipient)
```

### DIFF
--- a/crm/fcrm/doctype/crm_notification/crm_notification.py
+++ b/crm/fcrm/doctype/crm_notification/crm_notification.py
@@ -8,7 +8,8 @@ from frappe.model.document import Document
 
 class CRMNotification(Document):
 	def on_update(self):
-		frappe.publish_realtime("crm_notification")
+		if self.to_user:
+			frappe.publish_realtime("crm_notification", user= self.to_user)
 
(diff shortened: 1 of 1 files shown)


pu-raihan:
@shariquerik In a large-scale production environment with over 15 active users, assigning multiple users at once becomes highly challenging. Each assignment triggers a notification call for every user, leading to a significant number of API requests per user. As a result, the total server load multiplies based on the number of assignments and users, potentially causing performance issues.

This has actually happened with our client. So please fix ASAP

shariquerik (approved):

PR #658, opened 2025-03-12
feat: Onboarding & Help center

shariquerik:
Onboarding
<kbd>
image
</kbd>

Help center
<kbd>
image
</kbd>

Dependent on:
https://github.com/frappe/frappe-ui/pull/276
https://github.com/frappe/frappe/pull/31782

fixes: https://github.com/frappe/crm/issues/640


michellealva:
image

when the user has not started, maybe we should show "Start now" as the button label.

PR #693, opened 2025-03-23
fix: Default Arabic Font Family

abdulaziz-guru:
Currently, the Arabic default font is messy.  We need to use another appropriate font to resolve this issue.

### DIFF
--- a/frontend/src/index.css
+++ b/frontend/src/index.css
@@ -4,26 +4,32 @@
 @tailwind components;
 @tailwind utilities;
 
+@layer base {
+  * {
+    font-family: ui-sans-serif, system-ui, sans-serif, "Apple Color Emoji", "Segoe UI Emoji", Segoe UI Symbol, "Noto Color Emoji";
+  }
+}
+
 @layer components {
-	.prose-f {
-		@apply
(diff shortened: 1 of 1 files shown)


shariquerik:
@abdulaziz-guru cant go with this, we use InterVar font which is common across frappe apps, this changes will just override it. Not sure how it can be fixed

abdulaziz-guru:
> @abdulaziz-guru cant go with this, we use InterVar font which is common across frappe apps, this changes will just override it. Not sure how it can be fixed

The problem with "InterVar" is for some reason it look really messy in Arabic CRM pages. For example.

image

If I replace "InterVar" with any other font, problem will be resolved.

shariquerik:
@abdulaziz-guru This should fix your issue https://github.com/frappe/frappe-ui/pull/280

Will merge and release it soon

abdulaziz-guru:
> @abdulaziz-guru This should fix your issue [frappe/frappe-ui#280](https://github.com/frappe/frappe-ui/pull/280)
> 
> Already raised a fix pull latest and check

The font look readable now, much appreciated.

Issue #280, opened 2024-07-31
Add a full activity feed for contacts


the3dfox:
In the current implementation the activity feeds are very lead or deal specific, which means the whole communication only lives within the deal or lead. Sometimes you want to check on past communication where you don't specifically know the occasion or deal/lead you are looking for. Therefore a global or full activity feed should display all past activities with this specific contact, no matter the lead or deal linking, so you can quickly scroll through the feed to find the information you need.

I would suggest to add a global activity feed inside the contacts detail view, which could look like this
![image](https://github.com/user-attachments/assets/de611c62-142d-4d1b-9422-350c3ccb99b4)

Instead of just deals
![image](https://github.com/user-attachments/assets/2f379963-433f-44d0-8f0b-897f2a8a5537)

PR #702, opened 2025-04-01
feat: Add Activity tabs to contact

marcramser:
Enhance Contact Tab:
Add activity tabs to the "Contact" section to provide a clear overview of communications with existing customers/contacts.

Bildschirmfoto 2025-04-01 um 12 26 23


marcramser:
Solves https://github.com/frappe/crm/issues/280

shariquerik:
> @shariquerik what is your thoughts on that.

I am planning to make all view very similar with some common tabs like activity, email etc but it need some time. I am planning to do a big refactor using one component you will be able to build any doctype's list view and page view but it need many building blocks like better client script, almost all fieldtype support and there are many more small features. So lets park this for now.

PR #721, opened 2025-04-07
fix: restrict app in apps page if no access to FCRM module

shariquerik:


shariquerik:
@Mergifyio backport main

PR #726, opened 2025-04-08
fix: export logic

### LINKED ISSUE #673: Wrong checkbox "Export all" logic
When exporting a list of documents (all sections except "Notes" can be exported) to Excel or CSV, the "number of records" checkbox in the pop-up window does not work. If you click it or do not click it, in both cases all documents are downloaded. 

What is the idea? Maybe should appear additional filter after checkbox "Export All" is empty?

pratikb64:
Fixed export logic where while exporting data when user selects rows from list view only those rows are exported instead of all

Fixes #673

### DIFF
--- a/frontend/src/components/ViewControls.vue
+++ b/frontend/src/components/ViewControls.vue
@@ -545,6 +545,11 @@ function reload() {
 const showExportDialog = ref(false)
 const export_type = ref('Excel')
 const export_all = ref(false)
+const selectedRows = ref([])
+
+function handleSelectionsChange(selections) {
+  selectedRows.value = selections
+}
 
 async function exportRows() {
   let fields = JSON.stringify(list.value.data.columns.map((f) => f.key))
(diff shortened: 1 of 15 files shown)


pratikb64:
@mergifyio refresh

frontend/src/components/ViewControls.vue
```diff
@@ -560,7 +565,15 @@ async function exportRows() {
     page_length = list.value.data.total_count
   }
 
-  window.location.href = `/api/method/frappe.desk.reportview.export_query?file_format_type=${export_type.value}&title=${props.doctype}&doctype=${props.doctype}&fields=${fields}&filters=${filters}&order_by=${order_by}&page_length=${page_length}&start=0&view=Report&with_comment_count=1`
+  let url = `/api/method/frappe.desk.reportview.export_query?file_format_type=${export_type.value}&title=${props.doctype}&doctype=${props.doctype}&fields=${fields}&filters=${filters}&order_by=${order_by}&page_length=${page_length}&start=0&view=Report&with_comment_count=1`
+
+  // Add selected items parameter if rows are selected
+  if (selectedRows.value?.length && !export_all.value) {
+    url += `&selected_items=${JSON.stringify(selectedRows.value)}`
+  }
```
shariquerik, line 573: This will not work since selectedRows is a Set

shariquerik (changes requested):

frontend/src/components/ViewControls.vue
pratikb64: It works because at line 551 in updateSelections function it is getting converted to array

PR #845, opened 2025-05-21
feat: Users Settings Page

shariquerik:
Update Roles
image

Add Existing User
Screenshot 2025-06-20 at 7 02 40 PM

Invite New User
Screenshot 2025-06-20 at 7 03 16 PM

Screenshot 2025-06-20 at 7 04 08 PM

Change Password Modal
Screenshot 2025-06-20 at 7 07 59 PM


crm/fcrm/doctype/crm_agent/crm_agent.py
```diff
@@ -0,0 +1,58 @@
...
+	user_doc = frappe.get_doc("User", user)
+
+	if new_role == "Sales Manager":
+		user_doc.append_roles("Sales Manager", "System Manager")
+	if new_role == "Sales User":
+		user_doc.append_roles("Sales User")
+		if "Sales Manager" in frappe.get_roles(user_doc.name):
+			user_doc.remove_roles("Sales Manager", "System Manager")
+
+	user_doc.save()
```
kalungia, line 46: Thank you for the feature, but I’d like to raise a concern about how you have assigned roles to sales agents in `update_agent_role`.

Right now, you have assigned **Sales Manager**, **Sales User**, and even **System Manager** roles depending on the option selected. This approach has a few issues:

Sales agents can be internal (employees) or external (contractors), but we don’t treat them differently.

Giving roles like System Manager is risky — it gives full access.

Replacing or removing roles can affect internal users who already have other responsibilities.

Suggestion:

Let’s create a dedicated Sales Agent role.

For internal users, just add Sales Agent on top of existing roles.

For external users, only assign Sales Agent (with strict permissions).

This keeps things safer and easier to manage

kalungia (commented):
Thank you for the feature, but I’d like to raise a concern about how you have assigned roles to sales agents in `update_agent_role`.

Right now, you have assigned **Sales Manager**, **Sales User**, and even **System Manager** roles depending on the option selected. This approach has a few issues:

Sales agents can be internal (employees) or external (contractors), but we don’t treat them differently.

Giving roles like System Manager is risky — it gives full access.

Replacing or removing roles can affect internal users who already have other responsibilities.

Suggestion:

Let’s create a dedicated Sales Agent role.

For internal users, just add Sales Agent on top of existing roles.

For external users, only assign Sales Agent (with strict permissions).

This keeps things safer and easier to manage

shariquerik:
@Mergifyio refresh

shariquerik:
@Mergifyio rebase

crm/fcrm/doctype/crm_agent/crm_agent.py
kalungia: I wish sales agent was there also.

PR #1029, opened 2025-07-13
chore: update POT file

shariquerik:


mahsem:
@shariquerik POT file (added strings) sre not synced with crowdin for some time now

PR #1050, opened 2025-07-20
chore: update POT file

shariquerik:


barredterra (approved):

barredterra:
@shariquerik backporting POT changes doesn't make much sense – you're effectively overriding all translatable strings on `main-hotfix` with translatable strings from `develop`. Better trigger the "update POT file" workflow on the target branch directly.

Issue #730, opened 2025-04-10
Create and Make Call V1.40 Make Call not Working


sapl-santhosh:
In Version v1.40 the make call button not working. 

![Image](https://github.com/user-attachments/assets/d8a8bab8-ff34-4f63-a986-514ddf6ebf56)

shariquerik:
@sapl-santhosh Click on the primary button to trigger, dropdown is only there to select the action you want to perform.

PR #1070, opened 2025-07-24
refactor: Call Log Button

shariquerik:
Before:
<kbd>Screenshot 2025-07-24 at 1 13 03 PM

After:
<kbd>Screenshot 2025-07-24 at 1 12 22 PM

Fixes: https://github.com/frappe/crm/issues/730

### DIFF
--- a/frontend/src/components/MultiActionButton.vue
+++ b/frontend/src/components/MultiActionButton.vue
@@ -25,25 +25,22 @@
       </template>
     </Button>
     <Dropdown
-      v-show="showDropdown"
+      v-if="showDropdown"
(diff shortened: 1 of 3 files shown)


frontend/src/components/MultiActionButton.vue
```diff
@@ -57,13 +54,18 @@ const showDropdown = ref(props.options?.length > 1)
 const activeButton = ref(props.options?.[0] || {})
 
 const parsedOptions = computed(() => {
+  debugger
```
fproldan, line 57: Found a debugger here

PR #1133, opened 2025-08-11
fix(Org Modal): Setting loading false in case of Validation Error

wasim3357:
In this case user can add the details and save the form again instead of refreshing and then trying again for fresh.


wasim3357:
@shariquerik Can you please check this

PR #1136, opened 2025-08-12
Fix/details height behaviour

### LINKED ISSUE #986: UI: improve height behave of detail view to utilise max screen height
When I add a field to the detail view, the section is not scaling in height instead a scroll bar appears. Since there is still a lot of screen height left, I suggest making it responsive until the screen height is maxed out.

Image

YASH-SAWARKAR:
This solves the problem of  height behavior of detail view.

fixes #986

before 

image

Now 

image


shantanubilaw (approved):

shariquerik:
@YASH-SAWARKAR This is not the correct solution since the first section will take all the space, first section should only take the space if there is enough space after the section section.

Issue #796, opened 2025-05-02
Allow changing from while writing email


shariquerik:

atikrahmanbd:
Hi @shariquerik,

I hope you are doing well.

Let’s say we have two email accounts added: sales@business.xyz and person@business.xyz. It would be great if we could switch between email accounts while composing a new email or replying to existing messages.

Do you plan to work on this feature anytime soon? I truly appreciate the great work you’ve done!

Kind regards,
Atik

PR #1141, opened 2025-08-17
feat : Allows changing from while writing email

YASH-SAWARKAR:
Problem
- Email composer was missing a FROM field to select sender email
- Users couldn't choose which configured email account to send from

 Solution
- ✅ Added FROM field to email composer interface
- ✅ Created `SingleSelectEmailInput` component for single email selection
- ✅ Fetches available email accounts from `Email Account` doctype (with `enable_outgoing: 1`)
- ✅ Auto-selects first available email account by default

Let’s say we have two email accounts added: abc@gmail.com and xyz@gmail.com . Now we can manually select sender email  while composing a new mail or replying.
 
Closes : #796  #1078 

Before:
image

After:
image

image

### DIFF
--- a/frontend/src/components/Controls/SingleSelectEmailInput.vue
+++ b/frontend/src/components/Controls/SingleSelectEmailInput.vue
@@ -1,0 +1,215 @@
+<template>
+  <div>
+    <div class="flex flex-wrap gap-1">
+      <Button
+        v-if="selectedEmail"
+        :key="selectedEmail"
+        :label="selectedEmail"
+        theme="gray"
+        variant="subtle"
+        :class="{
+          'rounded bg-surface-white hover:!bg-surface-gray-1 focus-visible:ring-outline-gray-4':
+            variant === 'subtle',
+        }"
+      >
+        <template #suffix>
+          <FeatherIcon class="h-3.5" name="x" @click.stop="clearSelection" />
+        </template>
+      </Button>
+      <div class="flex-1">
+        <Combobox v-model="selectedValue" nullable>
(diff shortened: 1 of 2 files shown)


aquiveal:
You have no idea how timely this is! I was literally planning to work on this exact feature this weekend. Thank you so much for this PR!

YASH-SAWARKAR:
@shariquerik What are your thoughts on this? Does it need any changes?

shariquerik:
@YASH-SAWARKAR You are getting outgoing emails list which is not valid because there can be multiple outgoing emails but we should only show logged in user's emails check User doctype there is an 
Screenshot 2025-08-28 at 12 39 10 PM

You can check how this feature is implemented in desk for reference.

federicocalvo:
@shariquerik I don't know how they were using it, if this is a bug that appeared now as described here (https://github.com/frappe/crm/issues/1194) but this directly limits the most important use of the CRM, to be able to send mails having more than one account configured in the ERP, and the only way it works, is placing as default account the account you use in the CRM.

Unless there is something I'm missing I see several people interested in this PR maybe because of this.

YASH-SAWARKAR:
@federicocalvo I DM'd some queries to @shariquerik and I’m waiting for a reply. BTW, if someone has configured emails, they should be able to select the ‘From’ email of their choice (it solves this PR).

aquiveal:
@YASH-SAWARKAR The issue @shariquerik suggested earlier was that the user was given access to everyone's and all email accounts, which is not okay because any member could use each other's email accounts.

Issue #1189, opened 2025-08-26
Bug: Delete Confirmation Popup Not Closing After Deletion


YASH-SAWARKAR:
When deleting an Organization (or other records like Leads, Deals, etc.), the confirmation popup remains visible even after the record is successfully deleted. This leads to confusion as the deletion is already completed but the UI doesn’t reflect it immediately.

![Image](https://github.com/user-attachments/assets/894f99b4-f5cc-44d5-a1b1-457e18a65aab)

PR #1192, opened 2025-08-26
fixed popup deletion

patleYashpal:
It fixes problem related to abnormal behaviour of 2 deletion pop ups. First row selected pop up and second confirm delete pop up does not disappear after successful deletion.

Solves issue #1189

https://github.com/user-attachments/assets/e38c0f99-53fd-4878-8174-2b32a2596a9a


shariquerik:
@patleYashpal make sure you don't break indentation and add pre-commit to avoid missing it.

PR #1219, opened 2025-09-03
fix: address creation not returning to organization modal

aftabshaikh-agkiya:


aftabshaikh-agkiya:
https://github.com/frappe/crm/pull/1219

PR #1288, opened 2025-09-26
feat: FB Lead Sync first cut

NagariaHussain:


NagariaHussain:
That was the initial plan, but since lead syncing is very specific to CRM, we are going this way. Probably in the future.

pyproject.toml
```diff
@@ -61,4 +61,4 @@ typing-modules = ["frappe.types.DF"]
 [tool.ruff.format]
 quote-style = "double"
 indent-style = "tab"
-docstring-code-format = true
\ No newline at end of file
+# docstring-code-format = true
```
NagariaHussain:
```suggestion
docstring-code-format = true
```

Issue #1368, opened 2025-11-07
Delete button shown to users without delete permission


UmakanthKaspa:
The Delete button is currently visible to all users in the CRM portal, regardless of their permissions. Users who do not have delete access can still click the button and proceed through the delete confirmation modal, but they ultimately receive a PermissionError in the console.

**Steps to reproduce:**

- Log in as a user without delete access to CRM Leads.
- Click the Delete button.

PR #1372, opened 2025-11-07
Fix: Hide delete button when user lacks permission

UmakanthKaspa:
**issue**
The Delete button is currently visible to all users in the CRM portal, regardless of their permissions. Users who do not have delete access can still click the button and proceed through the delete confirmation modal, but they ultimately receive a PermissionError in the console.

**fix**
- Create useDocumentPermissions composable to check document permissions
- Add v-if check on delete button in Lead page
- Fetch permissions from frappe.client.get_doc_permissions API

**testing**
1. Login as user without delete permission for CRM Lead
2. Open any lead
3. Verify delete button is hidden

**before**
Button visible always 

**After**

https://github.com/user-attachments/assets/e82b123c-85d3-4749-a7a6-4e4edba1f4be

closes: #1368

### DIFF
--- a/crm/api/whatsapp.py
+++ b/crm/api/whatsapp.py
@@ -5,18 +5,19 @@ from frappe import _
 
 from crm.api.doc import get_assigned_users
 from crm.fcrm.doctype.crm_notification.crm_notification import notify_user
+from crm.integrations.api import get_contact_lead_or_deal_from_number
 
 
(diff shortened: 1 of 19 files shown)


frontend/src/pages/Contact.vue
shariquerik: canDelete should be part of useDocument

PR #1412, opened 2025-11-16
fix: surface deal value instead of annual revenue

longnguyen-osg:
## Summary
- Switch every Organization and Mobile Organization deals list to request and render `deal_value` (row formatter and column label), so the UI shows the actual deal amount instead of annual revenue.
- Update `CRMDeal.get_list_field_info()` to emit a “Deal Value” column backed by `deal_value`, keeping backend list metadata consistent with the frontend change.
- Extend the default “CRM Deal” side-panel and Data Fields layouts in `install.py` to include `deal_value`, so fresh installs expose the field automatically, and clean up unused lead-sync component typings.

## Testing
1. Open any Organization (desktop/mobile): the Deals table now shows a “Deal Value” column populated from `deal_value`.
2. Visit a Deal’s Activity → Data tab after syncing layouts or on a new site: the “Details” section includes the Deal Value input.


ps173:
some part of this was merged in #1581. Will look into the rest of the items and raise a separate PR

PR #1424, opened 2025-11-18
fix: Add permission check to /crm route

kitchanut:
## Description
This PR adds a security fix for the `/crm` route which currently allows any authenticated user to access the CRM application regardless of their roles or module access settings.

## Problem
Currently, `crm/www/crm.py` does not check user permissions in the `get_context()` function. This allows users to bypass:
- Module blocking (FCRM module)
- Role-based access control (Sales User/Manager roles)

## Solution
Added a call to `check_app_permission()` at the beginning of `get_context()` to verify:
- User has FCRM module enabled (not blocked)
- User has Sales User or Sales Manager role
- Throws `PermissionError` if conditions are not met

## Testing
1. Create a user without Sales User/Manager role
2. Block FCRM module for that user in User settings
3. Try to access `/crm` URL directly
4. User should see "You do not have permission to access CRM" error

## Impact
- Fixes security vulnerability where users can bypass access controls
- Ensures module blocking and role-based permissions are properly enforced

## Related Issues
Addresses security concern with unrestricted /crm route access

### DIFF
--- a/crm/www/crm.py
+++ b/crm/www/crm.py
@@ -32,7 +32,7 @@ def get_context():
 @frappe.whitelist(methods=["POST"], allow_guest=True)
 def get_context_for_dev():
 	if not frappe.conf.developer_mode:
-		frappe.throw("This method is only meant for developer mode")
+		frappe.throw(_("This method is only meant for developer mode"))
 	return get_boot()
 
 


crm/www/crm.py
shariquerik:
```suggestion
		frappe.throw(_("This method is only meant for developer mode"))
```

shariquerik (changes requested):

PR #1482, opened 2025-12-18
fix: use frappe.db.count instead of get_list for counting records

badal8381:
Summary

  - Replaced frappe.get_list() with count(*) field approach with frappe.db.count() for counting records
  - Fixes lead loading issues in the CRM portal

  Problem

  The CRM portal was failing to load leads due to the way record counts were being fetched. The previous implementation used:
  frappe.get_list(doctype, filters=filters, fields="count(*) as total_count")[0].total_count

  This approach can be unreliable and may fail in certain edge cases.

  Solution

  Replaced with the more robust and recommended Frappe API method:
  frappe.db.count(doctype, filters=filters)

  This change was applied in two locations within crm/api/doc.py:
  1. Kanban column count calculation (all_count)
  2. Total count for list view (total_count)

  Test Plan

  - Verify leads load correctly in CRM portal
  - Verify Kanban view displays correct counts
  - Verify list view pagination works correctly


badal8381:
Closing — the underlying v16 count-syntax bug was fixed upstream in 4e2c65b9 the day after this was opened, and the remaining diff would bypass the record-level permissions that the adjacent row queries apply, as the review bot correctly points out.

PR #1616, opened 2026-01-26
fix: only auto-fetch exchange rate if not already set by user

dominhhai:
Dear the maintainer team,

this PR improve the auto-fetch exchange rate by checking the inputed from user.

Thanks & best regards!


dominhhai:
@shariquerik 
Thanks for your response.
The bugs raised for these conditions

```python
if self.has_value_changed("currency") or not self.exchange_rate:
```

The changed currency will be trigged the `get_exchange_rate` function calling.
So, I added the logic to check whether user setting the `exchange_rate` before calling `get_exchange_rate` function.
Please checkout!

PR #1750, opened 2026-02-18
fix(notes): stop click propagation on overflow menu trigger in NoteArea

jeshani:
Fixes note card interaction where clicking ...  also trigger note edit modal.

delete


shariquerik:
@jeshani I dont think this issue exist anymore pull latest and check

PR #1791, opened 2026-02-27
fix(mobile): ensure status dropdown loads on initial render for Lead and Deal

### LINKED ISSUE #1790: Mobile: Lead and Deal status dropdown does not open on initial render
Describe the bug

In the mobile view of Lead and Deal pages, clicking on the status field does not show the list of available statuses during the initial render.
The current status label is visible, but when the user taps on it, the dropdown does not open or display any options.
After interacting with the page (for example, switching tabs or triggering a re-render), the status dropdown works correctly and displays the full list.
This issue only occurs in mobile view. Desktop view works as expected.

To Reproduce

1.Open Frappe CRM.
2.Switch to mobile view (responsive mode or mobile device).
3.Open any Lead or Deal document.
4.Tap on the status button at the top.
5.Observe that the dropdown list does not appear.
6.Switch to another tab (e.g., Activity or Details).
7.Tap on the status again.
8.Now the dropdown shows all available statuses.

Expected behavior

When clicking the status field in mobile view, the dropdown should immediately display the list of available statuses on the first render, without requiring additional interaction.

Additional context

Frappe CRM version: v1.59.0 (HEAD)

Affects:

MobileLead.vue
MobileDeal.vue

Desktop views are not affected.
Behavior suggests dropdown mounting before status options are properly initialized in mobile layout.

https://github.com/user-attachments/assets/bcf23f6a-7625-4f18-93d2-9bc0224e89e2

RahulK2202:
#1790 
Fixes an issue where the status dropdown in mobile Lead and Deal views did not open on initial render.

On first load in mobile layout, clicking the status field would not display the list of available statuses. After switching tabs or triggering a re-render, the dropdown would function correctly.

This was caused by evaluating `statusOptions()` inline in the template, which resulted in the dropdown mounting before options were properly resolved.

## Changes

- Replaced inline `statusOptions()` call in template with a computed `statuses` property.
- Bound `<Dropdown>` to `:options="statuses"` instead of calling the function directly.
- Applied fix to:
  - `MobileLead.vue`
  - `MobileDeal.vue`

## Result

- Status dropdown now opens correctly on first interaction.
- Consistent behavior between mobile and desktop views.
- No backend changes.

Fixes #1790

### DIFF
--- a/frontend/src/pages/MobileDeal.vue
+++ b/frontend/src/pages/MobileDeal.vue
@@ -10,17 +10,9 @@
       </Breadcrumbs>
       <div class="absolute right-0">
         <Dropdown
-          v-if="doc"
-          :options="
-            statusOptions(
-              'deal',
-              document.statuses?.length
-                ? document.statuses
-                : document._statuses,
-              triggerStatusChange,
-            )
-          "
-        >
+        v-if="doc && document.statuses"
+        :options="statuses"
+      >
           <template #default="{ open }">
             <Button
               v-if="doc.status"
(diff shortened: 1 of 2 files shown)


ps173:
Closing this as we were unable to replicate the bug and lack of verification from the authors end. Check #1790

PR #1817, opened 2026-03-04
feat(crm): enable table attachments in side panel and persist grid attachment actions

JanluPaulsenEpiUse:
Enable Table and Attach/Attach Image field types in CRM layout editors and side panel rendering.
Render table labels above table controls in side panel sections and support attachment upload/change/remove with icon-based actions.

Fix side-panel grid save flow so table row edits and attachment uploads are persisted.
Fix attachment “view” actions by using dedicated open handlers for reliable file opening.

image
image


ps173:
CRM doesn't ship with your added custom field "Attachments", hence there is no use case of adding this support in sidebar 

Also before directly raising a PR please raise a issue first!

PR #1890, opened 2026-03-23
fix: Create lead from incoming email from email account

shariquerik:
Screenshot 2026-03-23 at 12 14 12 PM


shariquerik:
> @shariquerik creates the initiative but does not record the email in the initiative (the content of the inquiry, so to speak)

Fixed: https://github.com/frappe/crm/pull/2002

PR #1999, opened 2026-04-06
refactor: Field layout editor

shariquerik:
This PR refactors the Field Layout Editor UI in Frappe CRM and significantly expands its capabilities — covering drag-and-drop behaviour, new field type support, file uploading, and automatic attachment cleanup.

**Key Changes**
1. Field Layout Editor UX Refactor
	- Drag-and-drop behaviour has been reworked for a cleaner experience.
	- Destructive actions (e.g. removing fields/sections) now show confirmation dialogs.
	- An empty-state UX is introduced to handle cases where no fields are present.

2. New Field Type Support
	- Three new field types are now supported in the layout editor: Attach, HTML, and Geolocation.
	- These are wired into the FieldLayout, SidePanel, and Grid renderers, so they render correctly wherever a layout is used.

3. File Upload Enhancements & Attachment Cleanup
	- File uploads now include the fieldname in the payload.
	- When a file is replaced (e.g. a new attachment uploaded to the same field), the old attachment is automatically cleaned up — both on the server side and the client side.


shariquerik:
@copilot summarises the PR in the description

PR #2008, opened 2026-04-09
fix: initialize FCRM Settings currency from site default after setup

### LINKED ISSUE #1967: FCRM Settings currency initializes to INR instead of site default currency after fresh setup
## Summary

On a fresh install of Frappe CRM, `FCRM Settings.currency` is initialized as `INR` even when the site default currency is `USD`.

This causes CRM dashboards and currency formatting to use the wrong currency immediately after setup.

## Environment

- Frappe `16.12.2`
- ERPNext `16.12.0`
- HRMS `16.4.5`
- Fresh site created in Docker

## Steps To Reproduce

1. Create a fresh site with Frappe, ERPNext, HRMS, and CRM installed.
2. Complete setup normally.
3. Open CRM settings or inspect the `FCRM Settings` single doctype.
4. Compare the CRM currency against the site defaults.

## Actual Result

- `Global Defaults.default_currency = USD`
- `System Settings.country = United States`
- `FCRM Settings.currency = INR`

CRM then displays currency values using INR instead of the site default currency.

## Expected Result

`FCRM Settings.currency` should inherit the site default currency, or at minimum match `Global Defaults.default_currency` after setup completes.

## Additional Notes

- This happened on a fresh site with no manual currency change in CRM.
- I could not find an explicit hardcoded `INR` assignment in the CRM codebase.
- A local workaround was to sync `FCRM Settings.currency` from `Global Defaults.default_currency` after setup completes.

## Upstream Investigation

- `frappe/frappe` owns the Desk and core setup wizard flow.
- `frappe/erpnext` extends the setup wizard and correctly writes the selected wizard currency into ERPNext defaults, including `Global Defaults.default_currency`.
- `frappe/crm` keeps its own separate currency in `FCRM Settings.currency`.
- I could not find code in CRM that reliably syncs `FCRM Settings.currency` from `Global Defaults.default_currency` after setup completes.

This makes the issue appear to be in CRM's setup/default-sync behavior rather than in Desk or ERPNext setup.

## Possible Root Cause

CRM appears not to reliably sync its own currency setting from the site defaults during or after setup.

## Workaround

Manually set `FCRM Settings.currency` to match `Global Defaults.default_currency`, or sync it during setup completion.

Tim1119:
## Problem
After a fresh install, `FCRM Settings.currency` is not initialized from the currency selected during the setup wizard. It remains empty or defaults to INR regardless of what was chosen.

## Root Cause
The `setup_wizard_complete` hook only called `create_demo_data` , nothing was reading the selected currency from `Global Defaults` and writing it  to `FCRM Settings`.

## Fix
Added `set_default_currency()` to the `setup_wizard_complete` hook. It reads `Global Defaults.default_currency` and sets it on `FCRM Settings.currency`  if not already configured.

## Testing
Fresh install with United States / USD selected in setup wizard → FCRM Settings correctly shows USD after setup completes.

Fixes #1967

### DIFF
--- a/crm/hooks.py
+++ b/crm/hooks.py
@@ -87,7 +87,10 @@ website_route_rules = [
 # Setup wizard
 # setup_wizard_requires = "assets/crm/js/setup_wizard.js"
 # setup_wizard_stages = "crm.setup.setup_wizard.setup_wizard.get_setup_stages"
-setup_wizard_complete = "crm.demo.api.create_demo_data"
+setup_wizard_complete = [
+    "crm.install.set_default_currency",
+    "crm.demo.api.create_demo_data",
+]
 # setup_wizard_test = "crm.setup.setup_wizard.test_setup_wizard.run_setup_wizard_test"
 
 # Installation
--- a/crm/install.py
+++ b/crm/install.py
@@ -31,6 +31,7 @@ def after_install(force=False):
 	create_default_manager_dashboard(force)
 	create_assignment_rule_custom_fields()
 	add_assignment_rule_property_setters()
+	set_default_currency()
 	frappe.db.commit()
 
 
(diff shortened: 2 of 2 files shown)


KappTech88 (approved):

ps173:
@Tim1119 please fix linter errors

Issue #1897, opened 2026-03-23
CRM socket.js uses common_site_config.json socketio_port instead of site-specific config


nilpatel42:
`apps/crm/frontend/src/socket.js` imports `socketio_port` from `common_site_config.json` at build time:

```js
import { socketio_port } from '../../../../sites/common_site_config.json'
```

In a multi-site bench setup, each site can have its own `socketio_port` in `site_config.json`. Frappe Desk correctly uses `window.socketio_port` which is injected per-site from `frappe.conf` via `base.html`. CRM bypasses this and always uses the common config value regardless of the site being served.

**To reproduce:**
1. Multi-site bench with different `socketio_port` in a site's `site_config.json` vs `common_site_config.json`
2. Access Frappe Desk → socketio connects on correct site port
3. Access Frappe CRM → socketio connects on common config port, WS fails

**Expected:** CRM should respect the per-site `socketio_port` the same way Frappe Desk does.

PR #2010, opened 2026-04-09
fix: use per-site socketio_port instead of common config

### LINKED ISSUE #1931: Auto-create a Lead when a WhatsApp message arrives from an unknown number
Right now when a WhatsApp message comes in from a number that has no matching Contact, Lead, or Deal, the message is saved but nothing else happens. An agent has to notice the unlinked message and manually create a lead from it, which is easy to miss.

For incoming emails the CRM already handles this automatically. The create_lead_from_incoming_email function in crm/utils/__init__.py creates a lead whenever an email arrives from an address that is not in the system, as long as the email account has that option enabled. It would be useful to have the same behaviour for WhatsApp.

The idea would be to add an optional setting (similar to the email one) that automatically creates a CRM Lead when an incoming WhatsApp message has no matching Contact or Lead. The lead name could default to the phone number until an agent fills in the details.

This would be most useful for teams using WhatsApp as a primary inbound channel where the first message from a prospect is often from a number that has never been seen before.

Happy to look into a PR for this if the approach sounds reasonable.

Tim1119:
## Problem
In a multi-site bench setup, `socket.js` imports `socketio_port` from `common_site_config.json` at build time. This means all sites use the same socketio port regardless of what is configured in each site's `site_config.json`, causing WebSocket connections to fail on sites with a different `socketio_port`.

## Root Cause
Two issues:
1. `socket.js` imports `socketio_port` from `common_site_config.json`  at build time — this is a static value shared across all sites
2. `crm/www/crm.py` never included `socketio_port` in the boot context, so `window.socketio_port` was always `undefined` in CRM's frontend

## Fix
1. Added `socketio_port` to `get_boot()` in `crm/www/crm.py` so it is  injected per-site at runtime via the boot context
2. Updated `socket.js` to use `window.socketio_port` instead of the build-time JSON import, matching how Frappe Desk handles this in `base.html`

## Testing
Multi-site bench with two sites:

- `opensource.localhost` — `socketio_port: 9000` → connects to **9000** ✅
opensource.localhost connecting to port 9000
*WebSocket successfully connects to port 9000 (matches site config)*

- `crm2.localhost` — `socketio_port: 9001` in `site_config.json` → attempts **9001** first ✅
crm2.localhost attempting port 9001
*CRM correctly reads per-site config and attempts port 9001 first. Connection refused because no socketio server is running on 9001 in this test bench — this confirms the per-site port is being read correctly. It then falls back to the available server on 9000.*

**Before fix:** `crm2.localhost` always connected directly to 9000 (common config), never attempting 9001

**After fix:** `crm2.localhost` correctly reads per-site `socketio_port: 9001` from `site_config.json`

Fixes #1931

### DIFF
--- a/crm/www/crm.py
+++ b/crm/www/crm.py
@@ -52,6 +52,7 @@ def get_boot():
 				"user": frappe.db.get_value("User", frappe.session.user, "time_zone")
 				or get_system_timezone(),
 			},
+			"socketio_port": frappe.conf.socketio_port or 9000,
 		}
 	)
 
--- a/frontend/src/socket.js
+++ b/frontend/src/socket.js
@@ -1,11 +1,10 @@
 import { io } from 'socket.io-client'
-import { socketio_port } from '../../../../sites/common_site_config.json'
 import { getCachedListResource, getCachedResource } from 'frappe-ui'
 
 export function initSocket() {
   let host = window.location.hostname
   let siteName = window.site_name
-  let port = window.location.port ? `:${socketio_port}` : ''
+  let port = window.location.port ? `:${window.socketio_port}` : ''
   let protocol = port ? 'http' : 'https'
   let url = `${protocol}://${host}${port}/${siteName}`
 


nilpatel42:
@Tim1119 

#1897

PR #2090, opened 2026-04-21
feat: sales hierarchy permissions for leads and deals

nextchamp-saqib:
Adds a `CRM Sales Hierarchy` tree doctype and a permission layer that restricts Lead and Deal visibility based on it.

When **Enable Sales Hierarchy Permissions** is turned on in FCRM Settings, a user can only see a record if:
- they own it, **or**
- it is owned by someone in their reporting subtree, **or**
- they are assigned to it via a ToDo, **or**
- a member of their subtree is assigned to it via a ToDo

Administrators and users with no hierarchy node always have full access.


shahzeelahmed:
continued in https://github.com/frappe/crm/pull/2120

Issue #2134, opened 2026-05-03
Auto creation of customers into ERPNext from Frappe CRM Error


GeoCommander:
**Describe the bug**
"Error while creating customer against Frappe CRM Deal"

Traceback (most recent call last):
  File "apps/erpnext/erpnext/crm/frappe_crm_api.py", line 162, in create_customer
    customer = frappe.get_doc({"doctype": "Customer", **customer_data}).insert(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "apps/frappe/frappe/model/document.py", line 309, in insert
    self.run_before_save_methods()
  File "apps/frappe/frappe/model/document.py", line 1143, in run_before_save_methods
    self.run_method("validate")
  File "apps/frappe/frappe/model/document.py", line 1014, in run_method
    out = Document.hook(fn)(self, *args, **kwargs)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "apps/frappe/frappe/model/document.py", line 1374, in composer
    return composed(self, method, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "apps/frappe/frappe/model/document.py", line 1356, in runner
    add_to_return_value(self, fn(self, *args, **kwargs))
                              ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "apps/frappe/frappe/model/document.py", line 1011, in fn
    return method_object(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "apps/erpnext/erpnext/selling/doctype/customer/customer.py", line 148, in validate
    self.validate_customer_group()
  File "apps/erpnext/erpnext/selling/doctype/customer/customer.py", line 334, in validate_customer_group
    frappe.throw(
  File "apps/frappe/frappe/__init__.py", line 615, in throw
    msgprint(
  File "apps/frappe/frappe/__init__.py", line 580, in msgprint
    _raise_exception()
  File "apps/frappe/frappe/__init__.py", line 531, in _raise_exception
    raise exc
frappe.exceptions.ValidationError: Cannot select a Group type Customer Group. Please select a non-group Customer Group.

**To Reproduce**
1. ERPNext and Frappe CRM on same site
2. Create Customer when deal is won

**Expected behavior**
[...]

PR #2136, opened 2026-05-04
Fix issue #2134

GeoCommander:
a temporary fix by entering "Individual" instead of "All Customer Groups" at erpnext_crm_settings.py


muchai:
This wil still fail, or force users to create a customer group, 'Individual'.
I think it should pick a value defined in the settings.

GeoCommander:
fair enough

PR #2182, opened 2026-05-15
feat: sync CRM Product with ERPNext Item master

shahzeelahmed:
## Summary

 This PR adds bidirectional product sync between Frappe CRM and ERPNext.
 When the integration is enabled, creating/updating/renaming/deleting a CRM Product mirrors changes to an ERPNext Item record, and vice versa.

  ## How it works

  We already have a settings doctype called `ERPNext CRM Settings`.
  Product sync can only run if both ERPNext & CRM are installed on the same site.
  If disabled => `CRM Product` and `Item` work separately, no syncing.

  If enabled:
  A "Sync Now" button is shown on the ERPNext settings page in CRM.

  > You need permission to write in both doctypes to perform this action.
  
 This is what happens when the button is pressed (it also runs automatically the first time the integration is enabled):
  
  A job is enqueued to sync the two masters.
  So let us say if

  ERPNext Item has:
  - ERP A
  - ERP B
  - Common

  CRM Product has:
  - CRM A
  - CRM B
  - Common

  After the job is finished both will have
  - ERP A
  - ERP B
  - Common
  - CRM A
  - CRM B

 Same code masters are linked, not duplicated. ERPNext is treated as source of truth for catalogue fields (price, image, description, disabled). CRM-side product names are preserved when both sides already have a name.

  > User Perms & Shared docs are also synced when the integration is enabled.
  
After the initial sync, ongoing edits flow both ways automatically: rename an Item → CRM Product renames; delete an Item → CRM Product is deleted; update price on either side → other side reflects it. If a linked Item is removed out of band, the CRM Product is unlinked and surfaced as a Sync Issue on the settings page.


shahzeelahmed:
tests failing due to recent framework changes in `add` method which removes `ignore_permissions` kwarg. will fix that in another PR. merging this for now

PR #2208, opened 2026-05-20
fix: add whatsapp template create button in popup

### LINKED ISSUE #2176: No create button in whatsapp template popup
### Description
If there is no template then I see a create template button in the whatsapp templates modal but if there are already existing templates then there is no visible create button. There should be a button to allow creation of template

# Screenshot
Image

ps173:
- Resolves #2176
- adds truncate in template name + title for accessibility

image


shahzeelahmed:
add screenshot

shariquerik (approved):

PR #2228, opened 2026-05-27
fix: resolve @me filters in export all records flow

### LINKED ISSUE #2227: Export All Records ignores @me filters in CRM Portal exports
### Description

When exporting records from the CRM Portal using the **Export All Records** option, filters containing the special `@me` value are not resolved before the export request is sent to the backend.

As a result, exports return blank/empty data even though records are visible correctly in the list view.

This issue specifically affects filters such as:

* `{ owner: "@me" }`
* `{ owner: ["=", "@me"] }`

The list view resolves `@me` correctly during normal data loading, but the export flow serializes filters directly without converting `@me` into the current logged-in user ID.

---

## Steps to Reproduce

1. Open CRM Portal list view
2. Apply a filter using `@me`

   * Example: `Owner = @me`
3. Verify records are visible in the list
4. Click **Export**
5. Select **Export All Records**
6. Download exported file

---

## Actual Result

The exported file is blank / contains no records because the backend receives unresolved `@me` values in the filters.

---

## Expected Result

`@me` should be resolved to the current logged-in user before sending export filters to the backend, so exported data matches the visible filtered records in the list view.

---

## Additional Context

Exporting **selected records** works correctly because that flow directly exports the selected document IDs and does not rely on resolving `@me` filters.

The issue only occurs in the **Export All Records** flow, where records are fetched entirely based on serialized filters.

---

## Root Cause

The export flow merges and serializes filters directly:

* `props.filters`
* `list.value.params.filters`

but does not transform `@me` into the active user ID before calling the export API.

---

## Proposed Fix

Resolve `@me` values before serializing export filters by replacing them with the current logged-in user ID.

This should support:

* Direct filters
* Operator-based filters

so export behavior remains consistent with the list view filtering logic.

Hemil-Sangani:
## Summary

Fixes #2227 by resolving `@me` filters before exporting records in the CRM Portal **Export Records** flow.

Previously, exports returned blank data when using **Export All Records** with filters containing the special `@me` value because the export request serialized filters directly without converting `@me` into the current logged-in user.

This issue affected filters such as:

* `{ owner: "@me" }`
* `{ owner: ["=", "@me"] }`

The list view itself worked correctly because `@me` was resolved during normal data loading, but the **Export All Records** flow skipped this transformation.

---

## Root Cause

The **Export Records** flow sent `@me` directly to the backend instead of resolving it to the current logged-in user.

Because of this, the backend could not match records correctly and returned empty export results.

---

## What Changed

* Added `@me` resolution logic before export filter serialization
* Replaces `@me` with the current logged-in user ID using `getUser().name`
* Supports:

  * Direct filter format
  * Operator-based filter format

Examples:

```js id="s6v78q"
{ owner: "@me" }
```

becomes:

```js id="0j5u8q"
{ owner: "user@example.com" }
```

and:

```js id="26l9mr"
{ owner: ["=", "@me"] }
```

becomes:

```js id="evn0ha"
{ owner: ["=", "user@example.com"] }
```

---

## Additional Context

Exporting **selected records** was already working correctly because that flow exports explicit selected document IDs instead of relying on filters.

The issue mostly affected the **Export All Records** flow, where data retrieval depends entirely on serialized filters.

### DIFF
--- a/frontend/src/components/ViewControls.vue
+++ b/frontend/src/components/ViewControls.vue
@@ -367,7 +367,7 @@ const props = defineProps({
 const { brand } = getSettings()
 const { $dialog, $socket } = globalStore()
 const { reload: reloadView, getDefaultView, getView } = viewsStore()
-const { isManager } = usersStore()
+const { isManager, getUser } = usersStore()
 const { organizations } = organizationsStore()
 
 const list = defineModel({ type: Object, default: () => ({}) })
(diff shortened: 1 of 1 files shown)


Hemil-Sangani:
Hi @shahzeelahmed Sir and @ps173 Sir, whenever you get a chance, could you please review this PR? I'd really appreciate your time and feedback. Thank you!

shahzeelahmed:
LGTM 👍

PR #2246, opened 2026-06-03
fix: render interactive WhatsApp messages with buttons

harshpwctech:
content_type == "interactive" rendered as empty bubbles because WhatsAppArea.vue had no matching render branch, and get_whatsapp_messages didn't return the buttons field. The fix adds an interactive render branch (with button list) and includes buttons in the API response


ps173:
Firstly, too much noise here, please thoroughly go through [contributing guide](https://github.com/frappe/frappe?tab=contributing-ov-file) before raising any PR. Also this feature will be catered in #2244

PR #2280, opened 2026-06-11
fix: incorrect contact in Lead to Deal conversion 

shahzeelahmed:


ps173:
Why was this merged ? Uniquely identifying leads using email opens door for a lot of problems as email itself is not a required field in CRM Lead. Hence the "mobile no" check was there in the first place. Please add a description

PR #2288, opened 2026-06-11
fix: data field tab

### LINKED ISSUE #2270: User should stay on same tab
**Describe the bug**
When the user do the change in any of the fields under different tabs, they are redirected to the Deal Details tab.
Instead, they should stay on the same tab.

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to 'data tab'
2. change any field value
3. click save

**Expected behaviour**

The behaviour should be sync which does not re-render the page or if re-render it than go back to the tab where it was before

**Screenshots**
Example : if I change "gender" in tab "2", it redirects me to the tab "1"

https://github.com/user-attachments/assets/00e26e9f-dd5e-490f-8981-7fb20a8ff0a4

**Desktop (please complete the following information):**
 - Version develop-crm

**Additional context**

yuanran32:
## Describe the bug

  When a user edits and saves a field under a non-default Data Fields tab, the form layout is re-
  rendered and the selected tab is reset to the first tab, such as Deal Details. This makes the
  user lose their current context after saving.

  ## Root cause

  `FieldLayout` kept the selected tab in local component state initialized to `0`. When the Data
  Fields layout was remounted or reloaded after save, that local state was recreated and the first
  tab became selected again.

  ## Changes

  - Lifted the Data Fields tab state to `Activities.vue` so it survives `DataFields` remounts.
  - Added `v-model:tabIndex` and `v-model:tabName` support to `FieldLayout`.
  - Restored the selected tab by stable tab `name`, which also works when tab labels are
  duplicated.
  - Persisted the selected Data Fields tab in `sessionStorage` per `doctype` and `docname`, so the
  tab can be restored after page remounts.
  - Added unit tests for tab persistence across layout remounts, DataFields remounts, tab reloads,
  and session storage restore.

  ## Expected behaviour after fix

  After editing and saving a field under another Data Fields tab, the user stays on the same tab
  instead of being redirected back to the first tab.

  ## Test plan

  - Added unit coverage in `frontend/tests/unit/fieldLayoutTabState.test.js`.
  - Run:

  ```bash
  cd frontend
  yarn test:run
```
closes #2270

### DIFF
--- a/frontend/src/components/Activities/Activities.vue
+++ b/frontend/src/components/Activities/Activities.vue
@@ -392,6 +392,8 @@
     </div>
     <div v-else-if="title == 'Data'" class="h-full flex flex-col px-3 sm:px-10">
       <DataFields
+        v-model:fieldLayoutTabIndex="fieldLayoutTabIndex"
+        v-model:fieldLayoutTabName="fieldLayoutTabName"
         :doctype="doctype"
         :docname="docname"
         @beforeSave="(data) => emit('beforeSave', data)"
@@ -531,6 +533,8 @@ const doc = computed(() => _document.doc || {})
 const reload_email = ref(false)
 const modalRef = ref(null)
 const showFilesUploader = ref(false)
+const fieldLayoutTabIndex = ref(0)
+const fieldLayoutTabName = ref('')
 
 const title = computed(() => props.tabs?.[tabIndex.value]?.name || 'Activity')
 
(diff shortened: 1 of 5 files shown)


yuanran32:
@mergifyio queue

yuanran32:
@ps173 @shahzeelahmed Gentle ping for this PR which fixes #2270.
All CI checks passed, and test cases are added to keep users on the same tab after saving.
Could you help review this PR when you have time? Thanks!

shahzeelahmed:
looks mostly correct. will test and merge

Issue #2271, opened 2026-06-09
Timestamp presentation and chronological direction of Activity/Emails


trufurs:
## Is your feature request related to a problem? Please describe.
I find the rigid timeline presentation in the activity feed partially restrictive. Depending on the context, "relative time" (e.g., 2 hours ago, 3 weeks ago) can be vague when trying to pinpoint exact touchpoints with a client. Conversely, forcing an absolute timestamp makes quick relative scanning harder. Additionally, the fixed chronological direction forces users into a single scrolling workflow, which might conflict with how individual sales reps prefer to read history (newest vs. oldest first).

## Describe the solution you'd like
I would like to introduce two user-level configuration preferences in the CRM settings to customize the Activity/Email Timeline:

Timestamp Format Toggle: Allow users to choose between Relative Time (e.g., 5 mins ago, 1 month ago) and Absolute Timestamp (e.g., June 9, 2026, 14:02).

Timeline Sorting Orientation: A toggle or dropdown allowing users to choose their preferred chronological order:

Descending: Newest activities/emails at the top (default).

Ascending: Oldest activities/emails at the top (ideal for reading a deal's history like a narrative from start to finish).

## Describe alternatives you've considered
Hovering: Relying on hovering over a relative timestamp to view the absolute time via a tooltip. However, this slows down productivity and is completely unusable on mobile/tablet devices.(already implemented)

Global system changes: Changing the core framework settings, but this should ideally be a user-level UI preference rather than a system-wide imposition.

## Additional context
These options would significantly improve UX for power users and teams migrating from other CRMs (like HubSpot or Salesforce) who are accustomed to specific timeline flows.

trufurs:
I have created a POC. Its [PR ](https://github.com/trufurs/crm/pull/3)

https://github.com/user-attachments/assets/2630adf3-65cc-40d3-b32d-a9682f09139b

trufurs:
@shahzeelahmed Any thought for improvement and progress on this? What ETA this can be merged and added?

ps173:
> [@shahzeelahmed](https://github.com/shahzeelahmed) Any thought for improvement and progress on this? What ETA this can be merged and added?

Minor nitpicky stuff only. We will eventually add this on page level as well once we figure single page layout doctype for the UI. ref #2065

ps173:
This has been released @trufurs

Issue #2065, opened 2026-04-16
Refactor contact and organization page


michellealva:
Similar to Lead and Deal page

PR #2290, opened 2026-06-12
feat: configurable activity timeline — timestamp format & sort order

trufurs:
## Summary
Adds two global (FCRM Settings) preferences for the activity timeline, set in
General Settings and applied for all users:
- Timestamp format: Relative (5 mins ago) or Exact date & time
- Sort order: Oldest First or Newest First (Activity / Comments / Emails / Calls)

The timestamp format also applies to the list views (Leads, Deals, Contacts,
Organizations, Tasks, Call Logs) and related deal/contact lists.

## Commits (setting → feature)
1. add timeline sort order setting
2. apply newest/oldest sort order in the activity timeline
3. add timeline timestamp format setting
4. show relative/exact timestamps in the activity timeline
5. apply timestamp format to list views

Issue https://github.com/frappe/crm/issues/2271

### DIFF
--- a/crm/fcrm/doctype/fcrm_settings/fcrm_settings.json
+++ b/crm/fcrm/doctype/fcrm_settings/fcrm_settings.json
@@ -15,6 +15,7 @@
   "auto_mark_replied_on_response",
   "auto_reopen_on_new_communication",
   "activity_timeline_section",
+  "crm_timeline_timestamp_format",
   "crm_timeline_sort_order",
   "currency_tab",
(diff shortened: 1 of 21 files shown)


shahzeelahmed:
LGTM

ps173 (approved):

ps173:
We should eventually extend this to add in the actual single page view as well in the page level setting. Will be done when we migrate page level views into layout doctype as well. ref #2065

PR #2310, opened 2026-06-18
fix: CRM Product pricing

shahzeelahmed:
### Summary

When the ERPNext integration is active, a CRM Deal product line should be priced the way ERPNext prices a Quotation line — resolved against the deal's own context (customer → price list → UOM → date) — instead of copying a single static "list price". This PR makes that change and fixes the underlying lookup bugs.
### Problem
  On a deal, the rate shown for a product didn't match the item's selling Item Price:

| Item | standard_rate | Item Price | Price List | UOM | Deal/CRM rate |
| :--- | :---: | :---: | :--- | :---: | :---: |
| CRM | 100 | 100 | Standard Selling | Nos | 100 |
| CRM-1001 | 90 | 1021 | _Test Price List 2 | Nos | 90 |

  1. UOM not passed: the deal-line lookup called get_item_price_rate(item_code) with no UOM, so ERPNext's get_item_price (IfNull(uom,'') IN ('', :uom)) matched only UOM-less prices and returned nothing, silently falling back to standard_rate.
  2. No deal context: pricing only consulted the default selling price list with no party, so a customer's price (e.g. on _Test Price List 2) could never surface, and the product standard_rate "reset" toward Item.standard_rate on every sync.


shahzeelahmed:
@mergify update

michellealva:
@shahzeelahmed please update the description.

shahzeelahmed:
@mergify update

PR #2314, opened 2026-06-18
Migrate to frappe-ui v1 (beta.11) + espresso v2 design system

ps173:
Migrates the CRM frontend onto frappe-ui **v1.0.0-beta.11** and the **espresso v2** design system. Reviewable as a stack of focused commits; the high-churn ones are mechanical codemods (noted below) so review can concentrate on the hand-written changes.

## What changed

- **Dependency upgrade** — pin frappe-ui to `v1.0.0-beta.11` (was lagging the v1 submodule), bump tiptap to v3 to match the editor's peer deps, and add `frappe-ui/editor` + `frappe-ui/internals` dev aliases so subpath imports resolve against the local submodule.
- **Design tokens (codemod, ~145 files)** — rename v0 semantic tokens to their v2 equivalents (`surface-white → surface-base`, `outline-gray-modals → outline-elevation-2`, etc.) and merge `text-size` + `font-weight` class pairs into the combined weight classes (`text-lg font-medium → text-lg-medium`). Purely mechanical.
- **Static icons → `lucide-*` mask utility (codemod)** — replace literal `<FeatherIcon name="...">` with the v1 `<span class="lucide-...">` utility, and prefix literal Button `icon`/`icon-left`/`icon-right` props with `lucide-`.
- **Dialog v1 flat-props API** — flatten `:options="{...}"` into discrete props, switch to `v-model:open`, and rename `#body-content`/`#body-title` slots to `#default`/`#title` across ~40 dialogs (incl. the programmatic `$dialog` helper).
- **Twilio settings** — swap the one deprecated frappe-ui `Autocomplete` for `Combobox`.
- **Tailwind content** — scan the linked `@framework/ui` source (`apps/frappe/ui/src`) so its utility/arbitrary-variant classes are emitted.

## Left on the back-compat path (intentional, follow-ups)

- **TextEditor — not migrated.** `TextEditorControl.vue` stays on the backward-compatible TextEditor API; only the unused `FeatherIcon` import was dropped. Migrating it to the new API is a deliberate follow-up.
- **Dynamic icons** — `:name`-bound `FeatherIcon` and `h(FeatherIcon)` render calls keep the FeatherIcon path (the `lucide-*` utility needs a literal class for Tailwind JIT); `FeatherIcon` stays globally registered for these.
- **`FormControl type="autocomplete"`** (LeadSyncSourceForm) — left as-is; its v-model holds the full option object, which `Combobox`'s value-only model would break.
- **Dialog `#body` full-overrides** — kept on the back-compat slot pending per-call-site review (`#body` is shared with Popover).

## TODO (follow-up)

- [ ] **Migrate the local `@/components/Icons/*.vue` SVG components to lucide.** These custom filled-SVG icons are still in use (`AppSidebar.vue` nav links / onboarding steps, plus ~50 files app-wide — e.g. `PhoneIcon` 20, `NoteIcon` 14, `TaskIcon` 12). Approach: replace with `~icons/lucide/*` **component imports** (matches the existing `LucideLayoutDashboard` usage; required because several consumers render via `<component :is>` / `markRaw`, which `lucide-*` class strings can't satisfy), then delete the components. Note: lucide is outline-style vs the current filled style, so this is a deliberate visual restyle.
  Proposed mapping: `InviteIcon→user-plus`, `EmailIcon→mail`, `CalendarIcon→calendar`, `PhoneIcon→phone`, `NotificationsIcon→bell`, `SquareAsterisk→square-asterisk`, `PinIcon→pin`, `DealsIcon→zap`, `LeadsIcon→users`, `ContactsIcon→square-user`, `OrganizationsIcon→building-2`, `NoteIcon→file-text`, `TaskIcon→square-check-big`, `CommentIcon→message-circle`, `StepsIcon→workflow`, `ConvertIcon→arrow-right-left`, `CollapseSidebar→panel-left-close`. `CRMLogo` stays (brand mark, no lucide equivalent).

## Notes for reviewers

- The notification components differ from `develop` by design-token renames only — this branch does **not** include the framework `NotificationPanel` experiment; the original CRM notification list is retained.


frontend/vite.config.js
```diff
@@ -130,6 +142,18 @@ function getAliases(config) {
...
+    // `.../src/index.ts/editor`. `internals` is pulled in by @framework/ui.
+    'frappe-ui/editor': path.resolve(
+      __dirname,
+      '../frappe-ui/src/molecules/editor/index.ts',
+    ),
+    'frappe-ui/editor-style.css': path.resolve(
+      __dirname,
+      '../frappe-ui/src/molecules/editor/style.css',
+    ),
+    'frappe-ui/internals': path.resolve(__dirname, '../frappe-ui/internals.ts'),
```
ps173, line 156: This will be removed in later release when we properly replace the current text editor with apt component.

PR #2316, opened 2026-06-20
fix: rename_Accounts_to_Account

mahsem:
Accounts in CRM collide with Accounts in ERPNext so change it to Account for ability to translate to other languages


ps173:
It is called accounts for a reason.

mahsem:
Yes but Accounting in ERPNext is also called Accounts
and have different meaning 
It is fine in english but in other languges is not
Maybe it should be called ”Email Accounts” instead

mahsem:
@ps173 maybe Accounts in ERPNext should be renamed to Accounting

ps173:
> Yes but Accounting in ERPNext is also called Accounts and have different meaning It is fine in english but in other languges is not Maybe it should be called ”Email Accounts” instead

Please raise a issue first with proof for this. Renaming it to account makes it weird from ux perspective as this has multiple email accounts. Your suggested Email Accounts makes sense hence show the translation in which this goes out of context and we can then think of renaming this. 

> @ps173 maybe Accounts in ERPNext should be renamed to Accounting

out of scope

PR #2360, opened 2026-06-25
fix: TypeError in add_forecasting_section on Python 3.14 (#2357)

### LINKED ISSUE #2357: Migration patch fails under Python 3.14: TypeError in add_forecasting_section_in_quick_entry
## Description

The migration patch `crm.patches.v1_0.add_forecasting_section_in_quick_entry` 
fails under Python 3.14 with a `TypeError`, causing `bench migrate` to abort 
and breaking all downstream functionality that depends on a completed migration.

## Environment

- **Frappe:** 16.24.0
- **ERPNext:** 16.25.0
- **CRM:** 1.74.1
- **Python:** 3.14
- **Hosting:** Frappe Cloud (private bench, no SSH access)

## Full Traceback
```
Executing crm.patches.v1_0.add_forecasting_section_in_quick_entry
Traceback (most recent call last):
File ".../frappe/migrate.py", line 143, in run_schema_updates
frappe.modules.patch_handler.run_all(
File ".../frappe/modules/patch_handler.py", line 76, in run_all
run_patch(patch)
File ".../frappe/modules/patch_handler.py", line 62, in run_patch
if not run_single(patchmodule=patch):
File ".../frappe/modules/patch_handler.py", line 152, in run_single
return execute_patch(patchmodule, method, methodargs)
File ".../frappe/modules/patch_handler.py", line 188, in execute_patch
_patch()
File ".../crm/patches/v1_0/add_forecasting_section_in_quick_entry.py", line 9, in execute
settings.add_forecasting_section("Quick Entry")
File ".../crm/fcrm/doctype/fcrm_settings/fcrm_settings.py", line 130, in add_forecasting_section
field
TypeError: cannot use 'dict' as a set element (unhashable type: 'dict')
```

## Root Cause

`add_forecasting_section` in `fcrm_settings.py` (line 130) performs a set 
operation on a list that contains `dict` objects. Dicts are not hashable and 
cannot be used as set elements. Python 3.14 enforces this strictly, causing 
the crash.

## Impact

Because `bench migrate` aborts at this patch, the `tabLetter Head` table never 
receives the new `letter_head_for` column introduced in Frappe 16.24.0 
(PR #38417). This causes **every print attempt on every voucher** to fail with:

`Field not permitted in query: letter_head_for
`

Printing is completely non-functional site-wide until migrate completes.

## Suggested Fix

In `fcrm_settings.py`, the set operation on line ~130 should compare dicts 
by a hashable key (e.g. `fieldname`) rather than putting the whole dict into 
a set. For example, replace:

```python
# Broken — dicts are not hashable
existing_fields = set(current_section.get("fields", []))
if field in existing_fields:
    ...
```

with:

```python
# Fixed — compare by fieldname string instead
existing_fieldnames = {
    f.get("fieldname") for f in current_section.get("fields", [])
}
if field.get("fieldname") in existing_fieldnames:
    ...
```

## Workaround (for Frappe Cloud users without SSH)

Insert the patch name into `__PatchLog` manually so it is skipped on the next 
migrate run. Users with SSH access can do:

```bash
bench --site <sitename> migrate --skip-failing
```

Frappe Cloud users without SSH have no self-service option and must rely on Frappe support to resolve this.

ArchitT:
## Problem

`add_forecasting_section` in `fcrm_settings.py` performs a set membership check using dict objects as set elements. Dicts are unhashable and cannot be set elements, causing:

`TypeError: cannot use 'dict' as a set element (unhashable type: 'dict')
`
This crashes `bench migrate` on Python 3.14, which prevents the `tabLetter Head` schema from being updated, breaking all PDF printing site-wide with `Field not permitted in query: letter_head_for`.

## Fix

Compare by `fieldname` (a hashable string) instead of by the full dict.

## Impact

This is a **production-blocking** issue for all Frappe Cloud sites running CRM + Frappe 16.24+ on Python 3.14. There is no self-service workaround for Frappe Cloud users (Query Console and System Console both block writes).

Closes #2357

### DIFF
--- a/crm/fcrm/doctype/fcrm_settings/fcrm_settings.py
+++ b/crm/fcrm/doctype/fcrm_settings/fcrm_settings.py
@@ -129,7 +129,7 @@ class FCRMSettings(Document):
 		# FIX: fields inside columns are dicts; extract "fieldname" string for set
 		# comparison instead of using dicts as set elements (unhashable on Python 3.14)
 		existing_fields = {
-			field.get("fieldname") if isinstance(field, dict) else field
+			(field.get("fieldname") or field.get("name")) if isinstance(field, dict) else field
 			for section in all_sections
 			for column in section.get("columns") or []
 			for field in column.get("fields") or []
--- a/crm/fcrm/doctype/fcrm_settings/fcrm_settings.py
+++ b/crm/fcrm/doctype/fcrm_settings/fcrm_settings.py
@@ -126,8 +126,10 @@ class FCRMSettings(Document):
 		)
 		if any(section.get("name") == "forecasted_sales_section" for section in all_sections):
 			return
+		# FIX: fields inside columns are dicts; extract "fieldname" string for set
+		# comparison instead of using dicts as set elements (unhashable on Python 3.14)
 		existing_fields = {
-			field
+			field.get("fieldname") if isinstance(field, dict) else field
 			for section in all_sections
 			for column in section.get("columns") or []
 			for field in column.get("fields") or []


shahzeelahmed:
only semantic commits check failing. non blocking

ArchitT:
This fix just missed the v1.75.0 cut by a few hours. Could we get a  v1.76.0 release soon? This is production-blocking so all PDF printing  is broken for affected sites until this lands in a tagged release.

shahzeelahmed:
@ArchitT will do a patch release

PR #2365, opened 2026-06-26
fix: rename_Accounts_to_Email_Accounts

mahsem:
Rename Accounts to Email Accounts  as Accounts have different context in ERPNext and is in line with Helpdesk naming


shahzeelahmed:
@mahsem looks good

Issue #2333, opened 2026-06-23
Allow defining custom dashboard charts instead of only the built-in ones


dpk404:
### Is your feature request related to a problem? Please describe.

I wanted to add a couple of charts to the CRM dashboard for some of our own fields and a custom doctype, and found there's no way to do it without editing CRM's own code.

The layout itself is editable (you can rearrange charts and it's saved on the CRM Dashboard doctype), but the set of charts you can pick from is fixed. Every chart is backed by a `get_<name>` function in `crm/api/dashboard.py`, and the backend will only render a chart if that function exists. So adding a new chart isn't a config change, it's a code change in core.

This is the bit that decides it, in `get_dashboard()`:

```python
for l in layout:
    method_name = f"get_{l['name']}"
    if hasattr(frappe.get_attr("crm.api.dashboard"), method_name):
        l["data"] = method(from_date, to_date, user)
    else:
        l["data"] = None
```

`get_chart()` does the same thing. If there's no matching `get_<something>`, the chart just comes back empty. The built-in ones (`get_total_leads`, `get_deals_by_source`, `get_sales_trend`, etc.) are all hardcoded here, and the default layout that lists them lives in `crm/fcrm/doctype/crm_dashboard/crm_dashboard.py`.

### Describe the solution you'd like

Something close to ERPNext's Dashboard Chart, but usable from inside the CRM dashboard: a way to define a chart from a data source (a report, a query, or a doctype with a group-by + count/sum/avg and some filters) and then place it on the dashboard alongside the built-in ones.

The frontend already renders number, axis (bar/line) and donut charts, so a custom chart could reuse those same types and the existing add/arrange UI. The piece that's missing is a way to register a chart and its data source without it having to be a `get_<name>` function in core.

### Describe alternatives you've considered

Overriding the whitelisted endpoints — pointing `get_dashboard`/`get_chart` at our own copies via `override_whitelisted_methods` so they also resolve our functions. It works, but we'd be re-implementing the core resolver and keeping it in sync every time CRM changes it, and only one app can own that override, so it breaks down as soon as two apps both want custom charts.

Frappe's own Dashboard Chart / Number Card doctypes already exist, but they aren't surfaced anywhere in the CRM UI, so they don't help here.

### Environment

- Frappe CRM 1.73.2
- Frappe Framework 16.22.0

PR #2416, opened 2026-07-06
feat: add chart options API and update AddChartModal component

trufurs:
## Summary

This PR refactors how chart options are loaded in the dashboard's **Add Chart** modal by moving chart definitions from the frontend to the backend.

Instead of maintaining hardcoded chart types and options in Vue components, the frontend now fetches them dynamically through a backend API. This makes the implementation easier to maintain and provides an extension point for apps to contribute their own dashboard charts in the future.

#### Backend

* Added a new `get_chart_options` API in `crm/api/dashboard.py`.
* Returns available chart types and their chart options grouped by type.
* Accessible to sales users.

#### Frontend

* Added a new dashboard composable to fetch chart options from the backend.
* Refactored `AddChartModal.vue` to consume the API instead of using hardcoded values.
* Simplified the chart selection logic by using a single dynamic control for all chart types.
* Updated the selected chart handling to work with dynamically loaded options.

### Benefits

* Removes hardcoded chart definitions from the frontend.
* Centralizes chart configuration in one place.
* Makes adding or modifying dashboard charts significantly easier.
* Provides a foundation for allowing custom apps to contribute dashboard charts, enabling features such as embedding CRM Dashboard charts created outside the core app (see #2333).

### After Change

*(working as expected)*

image

### Error Handling

*(API function intentionally renamed to verify error state)*

image

### POC

A proof of concept demonstrating how another feature can override `get_chart_options` to expose additional dashboard charts in CRM.

image

**Related:** frappe/crm#2333


frontend/src/composables/dashboard.js
```diff
@@ -0,0 +1,17 @@
...
+export const chartOptionsByType = ref({})
+
+export const chartOptions = createResource({
+  url: 'crm.api.dashboard.get_chart_options',
+  cache: 'crm-dashboard-chart-options',
+  auto: true,
+  onSuccess: (data = {}) => {
+    const { chart_types = [], ...optionsByType } = data
+    chartTypes.value = chart_types
+    chartOptionsByType.value = optionsByType
```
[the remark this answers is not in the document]
trufurs: Should i add Error toast message? As Intentional. The endpoint has no failure path independent of the rest of the dashboard (same auth, same server). A JS fallback would re-introduce the hardcoded arrays this PR removes and create a second source of truth. If the fetch fails, everything else on the dashboard already fails first.

shahzeelahmed:
looks good, could you please add some screenshots? @trufurs

trufurs:
@shahzeelahmed Please look into this. If you'd like, I can also create a new PR for adding charts and cards from Frappe Number Cards and Dashboard Charts to render on the CRM Dashboard UI.

This PR lets you override the fields in chart options, allowing developers to use it to show their own custom charts as well.

trufurs:
@shahzeelahmed  added screenshots, when this can be released ?

trufurs:
@shahzeelahmed  Please check when it can released as it will give better solution to for dashboard modifications for developers

trufurs:
@shahzeelahmed any updates on this ?

shahzeelahmed:
@trufurs haven't tested it out yet. for configuring, how can we make sure we're always passing correct params to axis (line,bar,etc) and non axis charts (donut, pie)? we need to make sure that edge cases are handled properly

trufurs:
@shahzeelahmed This PR only moves the hard coded variables to backend so it can be modified. 
if any developer further intergrate  `Frappe Number Cards and Dashboard Charts ` there number charts number only and Dashboard charts have type as donut and axis charts with correct data.
with custom overrides we can do that as a POC shown above.

trufurs:
@shahzeelahmed just a gentle follow up!!

frontend/src/components/Dashboard/AddChartModal.vue
```diff
@@ -1,6 +1,10 @@
 <template>
-  <Dialog v-model:open="show" :title="__('Add Chart')" @close="show = false">
-    <template #default>
+  <Dialog
```
shahzeelahmed: This reverts the Dialog to the deprecated `v-model` + `:options` + `#body-content` API. frappe-ui's `Dialog.vue` fires `warnDeprecated` for both the `options` prop and the `body-content` slot, so every open of this dialog logs two deprecation warnings.

46 other call sites in this app use `v-model:open` + `:title` + `#default`. The branch looks like it predates the frappe-ui v1 migration (#2314), so merging this silently re-reverts that file.

frontend/src/components/Dashboard/AddChartModal.vue
```diff
@@ -60,48 +51,20 @@ const toDate = inject('toDate', ref(''))
...
-const axisCharts = [
-  { label: __('Sales Trend'), value: 'sales_trend' },
-  { label: __('Forecasted Revenue'), value: 'forecasted_revenue' },
-  { label: __('Funnel Conversion'), value: 'funnel_conversion' },
-  { label: __('Deals by Ongoing & Won Stage'), value: 'deals_by_stage_axis' },
-  { label: __('Lost Deal Reasons'), value: 'lost_deal_reasons' },
-  { label: __('Deals by Territory'), value: 'deals_by_territory' },
-  { label: __('Deals by Salesperson'), value: 'deals_by_salesperson' },
-]
+const selectedChart = reactive<Record<string, string>>({
```
shahzeelahmed: `selectedChart` still hardcodes `sales_trend` / `deals_by_stage_donut` on the frontend, which is exactly what the PR is trying to move server-side.

If an app overrides `get_chart_options` and drops those values, the select renders its placeholder (the model value matches no option) but the model still holds the stale value. The user clicks **Add** without touching the select and a chart the override never offered gets requested and added.

Deriving the default from `currentOptions[0].value` in a watcher on `chartType` would fix this and remove the last hardcoded chart names from the component.

frontend/src/components/Dashboard/AddChartModal.vue
```diff
@@ -117,12 +80,7 @@ async function addChart() {
 }
 
 async function getChart(type: string) {
-  let name =
-    type == 'number_chart'
-      ? numberChart.value
-      : type == 'axis_chart'
-        ? axisChart.value
-        : donutChart.value
+  const name = selectedChart[type]
```
shahzeelahmed, line 99: There is no default for `number_chart`, so `selectedChart['number_chart']` is `undefined`.

Pick Chart Type = Number Chart, leave the second select on its placeholder, click **Add**: `get_chart` is called with `name=undefined`, the backend finds no `get_undefined` and returns `{"error": "Invalid chart name"}` — and `onSuccess` pushes it anyway as a tile named `undefined` with layout key `undefined_xxxx`, which then gets persisted into the saved layout.

Either seed the default from `currentOptions[0]` or disable **Add** while no option is selected (and skip the push when the response carries an `error`).

crm/api/dashboard.py
```diff
@@ -88,6 +88,45 @@ def get_chart(
 		return {"error": _("Invalid chart name")}
 
 
+@frappe.whitelist()
+@sales_user_only
+def get_chart_options():
```
shahzeelahmed: Fitness gap: an app that overrides `get_chart_options` still can't ship a chart that actually renders.

- `get_chart` resolves data only via `getattr(frappe.get_attr("crm.api.dashboard"), f"get_{name}")`, so a contributed option value has no resolver and returns `{"error": ...}`.
- Adding a new chart *type* is worse: `DashboardItem.vue` hardcodes `number_chart` / `axis_chart` / `donut_chart` in a `v-if` chain, so an unknown type renders nothing.
- `override_whitelisted_methods` is all-or-nothing: one app wins, and it has to restate the entire core list.

A `frappe.get_hooks("crm_dashboard_charts")` merge here (with the hook also supplying the data resolver) would make this composable across apps, which is what #2333 needs.

crm/api/dashboard.py
```diff
@@ -88,6 +88,45 @@ def get_chart(
 		return {"error": _("Invalid chart name")}
 
 
+@frappe.whitelist()
```
shahzeelahmed: This PR builds a canonical server-side allow-list of chart names and then doesn't use it — `get_chart` still dispatches on arbitrary caller input via `hasattr`/`getattr` over the module namespace.

Concretely, this function adds a new reachable target: any sales user can call `crm.api.dashboard.get_chart` with `name="chart_options"`, `hasattr(module, "get_chart_options")` is now `True`, and it gets called as `get_chart_options(from_date, to_date, user)` → `TypeError: takes 0 positional arguments but 3 were given`, i.e. a 500 with a traceback in the error log.

One-line fix: in `get_chart`, reject `name` unless it appears in the flattened values of `get_chart_options()`.

frontend/src/components/Dashboard/AddChartModal.vue
trufurs: i have updated the code with a hook based input , please check! and re review

shahzeelahmed (approved):

Issue #2430, opened 2026-07-09
feat: tint Kanban column background with its indicator color


palamars:
## Feature request

Right now a Kanban column's color (`column.column.color`, set via the color picker on each column) only tints the small indicator dot next to the column title — it has no effect on the column itself.

Frappe's classic Desk Kanban board (used in plain ERPNext) already colors the whole column background from the same kind of indicator value:

```html
<!-- frappe/frappe/public/js/frappe/views/kanban/kanban_column.html -->
<div class="kanban-column" style="background-color: var(--bg-{{indicator}});">
```

CRM's Vue rewrite of the Kanban view never carried this over, so today picking a column color has almost no visible effect — the board stays visually flat even with per-column colors assigned. Tinting the column background makes the board meaningfully easier to scan at a glance (matches the older Desk behavior, makes the existing color picker actually useful).

## Proposed implementation

Add a `parseBgColor()` helper next to the existing `parseColor()` in `frontend/src/utils/index.js` and apply it to the column container in `KanbanView.vue`, using the Tailwind `!bg-` pattern already covered by the project's safelist (`safelist: [{ pattern: /!(text|bg)-/, ... }]` in `tailwind.config.js`) — no config changes needed.

```js
export function parseBgColor(color) {
  if (!color) return ''
  if (color == 'black') return '!bg-gray-100'
  return `!bg-${color}-50`
}
```

I have a working implementation on a fork, happy to open a PR if this direction looks good — flagging as a feature/design proposal first since it's a visual change (rather than a pure bugfix) and might warrant discussion on shade/opacity before merging.

shahzeelahmed:
added explanation in the pr https://github.com/frappe/crm/pull/2433

PR #2433, opened 2026-07-09
feat: tint Kanban column background with the column's color

palamars:
## Summary

Proposal from #2430 — a column's color (set via the existing color picker on each Kanban column) currently only tints the small indicator dot next to the column title. This tints the whole column background too, mirroring how the classic Desk Kanban board (`frappe/frappe/public/js/frappe/views/kanban/kanban_column.html`) already colors its column background from the same kind of indicator value.

Adds `parseBgColor()` next to the existing `parseColor()` in `frontend/src/utils/index.js` and applies it to the column container in `KanbanView.vue`, using the `!bg-` Tailwind pattern already covered by the project's safelist (`safelist: [{ pattern: /!(text|bg)-/, ... }]`) — no config changes needed.

```js
export function parseBgColor(color) {
  if (!color) return ''
  if (color == 'black') return '!bg-gray-100'
  return `!bg-${color}-50`
}
```

Flagging this as a feature/design proposal rather than a plain bugfix — happy to adjust the shade/opacity or scope it differently if maintainers have a preference, this is meant as a starting point for discussion with a working implementation attached.

(Re-opened against `develop` — the original PR targeted `main` and was auto-closed by Mergify. Also restores `hover:bg-surface-gray-2` on the column container, dropped by mistake in the first version and caught by the greptile-apps review — it was silently killing the hover highlight on columns with no color assigned.)

## Test plan
- [x] Verified live on a deployed CRM instance: columns now show a light background tint matching their assigned color, indicator dot unchanged
- [x] No behavior change for columns without an explicit color (falls back to no tint, same as before)
- [x] Hover highlight confirmed working again on uncolored columns

Relates to #2430

### DIFF
--- a/frontend/src/components/Kanban/KanbanView.vue
+++ b/frontend/src/components/Kanban/KanbanView.vue
@@ -11,7 +11,7 @@
       <template #item="{ element: column }">
         <div
           v-if="!column.column.delete"
-          class="flex flex-col gap-2.5 min-w-72 w-72 rounded-lg p-2.5"
+          class="flex flex-col gap-2.5 min-w-72 w-72 hover:bg-surface-gray-2 rounded-lg p-2.5"
           :class="parseBgColor(column.column.color)"
         >
           <div class="flex gap-2 items-center group justify-between">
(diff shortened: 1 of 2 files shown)


shahzeelahmed:
@palamars  we don't want the kanban column backgrounds to have different colors and its not in accordance with the espresso design system. refer to this page https://ui.frappe.io/recipes/demo/deals-desktop

PR #2437, opened 2026-07-09
fix: SidePanelLayout missing "Create New" button for Link fields

### LINKED ISSUE #2423: SidePanelLayout missing auto-create for Link fields
## Description

`SidePanelLayout.vue` passes `:onCreate="field.create"` to the `<Link>` component (line 194), but never auto-assigns `field.create` for Link fields. This means the "Create New" button in Link field dropdowns only appears when `field.create` is explicitly set by the parent page (e.g., the `address` field in `Organization.vue`).

In contrast, `Field.vue` (used by `FieldLayout` in modals like `OrganizationModal`) has auto-create logic that works for **all** Link fields:

```js
// Field.vue — auto-create for every Link field (non-User)
if (field.fieldtype === 'Link' && field.options !== 'User') {
  if (!field.create) {
    field.create = (value, close) => {
      const callback = (d) => {
        if (d) fieldChange(d.name, field)
      }
      createDocument(field.options, value, close, callback)
    }
  }
}
```

`SidePanelLayout.vue` has no equivalent logic.

## Steps to reproduce

1. Add a custom Link field to CRM Organization (e.g., linking to a custom DocType)
2. Open an existing organization (sidepanel view, rendered by `SidePanelLayout`)
3. Click the Link field dropdown → **no "Create New" button**
4. Now open the "New Organization" modal (rendered by `FieldLayout` → `Field.vue`) → **"Create New" button IS shown** for the same field

## Expected behavior

"Create New" button should appear in Link field dropdowns in **both** the sidepanel view and the modal, consistently.

## Affected component

`frontend/src/components/SidePanelLayout.vue` — needs the same auto-create logic that `frontend/src/components/FieldLayout/Field.vue` already has.

## Environment

- FCRM v1.76.0
- Frappe v16.25.0

georgekimaninjoroge:
Fixes #2423

Link fields in the sidepanel never showed the "Create New" button 
because SidePanelLayout.vue didn't set up field.create. Field.vue 
already handles this — I copied the same pattern over.

Added the auto-create logic to parsedField() and imported 
createDocument from @/composables/document.


RitvikSardana:
Show SS

georgekimaninjoroge:
Picked this up from @andropl's report in #2423,  the reproduction steps were detailed enough to trace the issue without running it locally. The fix is a straight copy of the pattern Field.vue already uses for the same case, so I'm fairly confident it's correct

georgekimaninjoroge:
@RitvikSardana set up locally and confirmed it. before the fix the dropdown has no "Create New" option, after it shows up correctly. screenshots attached.
fix
bug

PR #2448, opened 2026-07-12
fix: view menu (...) button unclickable/unstable due to zero-width hover reveal

palamars:
## Problem

The per-view context menu trigger (Duplicate / Edit / Make Public / Delete) in the views dropdown had `width: 0` and `pointer-events: none` at rest, only becoming clickable on parent `:hover`:

```
class="group-hover:!w-auto !w-0 opacity-0 group-hover:opacity-100 pointer-events-none group-hover:pointer-events-auto"
```

Because the width itself animates `0 -> auto` on hover, the button's hit-box shifts position at the exact moment it appears. A real click that arrives in the same motion as the hover (normal fast mouse movement, not a deliberate hover-then-click-separately gesture) lands on the pre-shift position — either empty space, or the row's own click handler (which switches views) — instead of opening the menu. No console error, so it reads as "randomly doesn't work."

## Fix

Never gate a hover-reveal control's clickability on `width`/`pointer-events` changing together with opacity — keep the box always at its final size and always interactive, and hide it with opacity alone:

```diff
- class="group-hover:!w-auto !w-0 opacity-0 group-hover:opacity-100 pointer-events-none group-hover:pointer-events-auto"
+ class="opacity-0 group-hover:opacity-100"
```

Verified via `getComputedStyle` before/after (width `0px` -> `28px` constant, `pointer-events` `none` -> `auto` constant, only opacity still toggles) and a real (non-synthetic) click with no prior hover dispatch — opens reliably every time post-fix.

### DIFF
--- a/frontend/src/components/ViewBreadcrumbs.vue
+++ b/frontend/src/components/ViewBreadcrumbs.vue
@@ -34,52 +34,28 @@
           </template>
         </Button>
       </template>
-      <template #item="{ item, close }">
-        <button
-          class="group flex text-ink-gray-6 gap-4 h-7 w-full justify-between items-center rounded px-2 text-base hover:bg-surface-gray-3"
-          @click="item.onClick"
-        >
-          <div class="flex items-center">
-            <FeatherIcon
-              v-if="item.icon && typeof item.icon === 'string'"
-              :name="item.icon"
-              class="mr-2 h-4 w-4 flex-shrink-0 text-ink-gray-7"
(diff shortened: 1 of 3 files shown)


palamars:
Closing — this fix is superseded by #2450, which independently refactored this same button to a fixed-size, opacity-only hover reveal (`!size-5 opacity-0 group-hover:opacity-100`) as part of switching to frappe-ui's built-in selection handler. No remaining diff against `develop`.

PR #2454, opened 2026-07-13
feat: add ERPNext Item<> CRM Product sync toggle

shahzeelahmed:
Added a toggle for users who want to opt out of Item and Product sync

image


shahzeelahmed:
@greptileai review resolved changes and update the score

PR #2467, opened 2026-07-13
fix: preserve custom favicon in page metadata

gabrielagenciasegundo:
## Summary

- Preserve the configured brand favicon when page metadata changes.
- Apply the existing `FCRM Settings` brand value to Dashboard, Data Import, and Persona Form.

## Root cause

These pages called `usePageMeta` with a title but without an icon. The page metadata utility then restored the static default Frappe favicon, replacing the favicon configured in CRM branding after navigation.

## Impact

The browser tab now keeps the configured CRM favicon while navigating through these pages.

## Validation

- Prettier check passed for all three changed Vue files.
- `yarn test:run` passed with 7 test files and 128 tests.
- `git diff --check` passed.


shahzeelahmed:
please raise issues for these PRs first

PR #2474, opened 2026-07-14
feat: Web Form Builder in Frontend

### LINKED ISSUE #2063: Lead creation via in-built web form (via frontend)

michellealva:
Closes https://github.com/frappe/crm/issues/2063

## Web Forms for CRM

This adds a built-in form builder to CRM, so you can create public forms that collect leads and deals directly from your website.

### What it does

You'll find it under Settings → Customization → Forms. From there you can:

- Build a form visually. Pick which fields to show, arrange them into sections and columns, and drag them around to reorder. A searchable dropdown lets you add fields quickly.
- Connect it to your CRM. Each form maps to either a Lead or a Deal. When someone submits, a new record is created automatically and tagged with the source "Web Form."
- Handle deals smartly. For deal forms, the matching Organization and primary Contact are created and linked for you, the same way CRM does it internally.
- Publish and share. Turn a form on or off with a single toggle. Share it with a direct link or drop it into any website using an embed snippet.
- Embed anywhere. By default, a form only shows on your own site, but you can whitelist other domains so it works when embedded elsewhere.
- Preview drafts safely. While a form is still a draft, you can preview and test-submit it without creating real records.
- Redirect after submit. Optionally send people to a thank-you page after they submit, with a short countdown and a success message first.

### Demo

https://github.com/user-attachments/assets/1491ee31-2e1d-4482-92ea-86670dcbb1d6

### How it's built
Rather than reinventing form storage and validation, this uses Frappe's existing Web Form engine under the hood, while keeping a custom, polished builder interface and public page on top. Public submissions go through the framework's standard, validated path, so there's no extra public endpoint to worry about from a security standpoint. Only users with the CRM Manager role can create or edit forms.

### Testing
Includes a suite of tests covering the field picker, form creation, publishing rules, guest submissions, draft previews, and permission checks.

Docs: https://docs.frappe.io/crm/capturing-leads/web-form


crm/www/crm_form.py
```diff
@@ -0,0 +1,99 @@
+# Copyright (c) 2026, Frappe Technologies Pvt. Ltd. and contributors
+# For license information, please see license.txt
+
+import frappe
+
+no_cache = 1
+
+
+ALLOWED_DOCTYPES = ("CRM Lead", "CRM Deal")
```
ps173: can save this in boot with proper name

crm/hooks.py
```diff
@@ -180,10 +178,14 @@
...
 	},
 	"CRM Deal": {
+		"before_insert": ["crm.api.form.enrich_form_submission"],
 		"on_update": [
 			"crm.fcrm.doctype.erpnext_crm_settings.erpnext_crm_settings.create_customer_in_erpnext"
 		],
 	},
+	"CRM Lead": {
+		"before_insert": ["crm.api.form.enrich_form_submission"],
+	},
```
ps173: instead of hooks, do it inside the respective controllers

frontend/src/components/Settings/Forms/fieldTypeIcon.js
```diff
@@ -0,0 +1,56 @@
...
+    case 'Text Editor':
+    case 'HTML Editor':
+    case 'Markdown Editor':
+      return LucideAlignLeft
+    case 'Link':
+      return LucideLink
+    default:
+      return LucideType
+  }
+}
```
ps173, line 52: not a review but a note that we should probably extract this somewhere to be used in other apps in future

ps173 (commented):
also address greptile reviews as well

michellealva:
Addressed in 7926c5d7:

**greptile — hidden-defaults spoofing (`crm/api/form.py`)** — `_apply_hidden_defaults` now resolves the `web_form` param and bails unless it's a CRM form whose `doc_type` matches the record being inserted (single `db.get_value(..., as_dict=True)`), so a submission can't pull in another form's / another doctype's defaults. Added a test that inserts a Lead while pointing `web_form` at a Deal form and asserts the Deal's Status default doesn't land.

**greptile — `doc.set("web_form_fields", [])` wipes layout** — now guarded on `fields is not None`, so an update that omits `fields` (settings-only save / direct API call) no longer clears the layout. Added a test.

**@ps173 — enrichment in controllers, not hooks** — moved `enrich_form_submission` out of `doc_events` into the `before_insert` of the CRM Lead and CRM Deal controllers; dropped the hooks entries.

**@ps173 — `ALLOWED_DOCTYPES` in `crm_form.py`** — now imported from `crm.api.form` instead of redefined. (Let me know if by "save in boot" you meant exposing it via `boot` for the client)

**@ps173 — `fieldTypeIcon.js`** — noted as a good future extraction for reuse across apps; left as-is for this PR.

All 14 server tests pass.

PR #2477, opened 2026-07-15
fix(list-view): refetch cached list resource on remount

ps173:
The list resource was cached by [doctype, view, viewType] but never set auto: true, so navigating back to a list after editing something elsewhere (e.g. assigned_to) rendered the stale cached data with no refetch guarantee — the only correction path was a manual 100ms debounced reload on mount. frappe-ui's createResource already forces a reload on cache-hit when auto is true, so use that instead, consistent with the assignees cache in data/document.js.

### DIFF
--- a/frontend/src/components/Controls/GridRowFieldsModal.vue
+++ b/frontend/src/components/Controls/GridRowFieldsModal.vue
@@ -45,10 +45,9 @@
 <script setup>
 import FieldLayout from '@/components/FieldLayout/FieldLayout.vue'
 import FieldLayoutEditor from '@/components/FieldLayoutEditor.vue'
-import { useDebounceFn } from '@vueuse/core'
 import { useTelemetry } from 'frappe-ui/frappe'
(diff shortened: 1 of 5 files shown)


ps173:
Note on the commit history: this branch originally also applied the same `auto: true` fix to the four layout builder modals (`QuickEntryModal`, `SidePanelModal`, `DataFieldsModal`, `GridRowFieldsModal`) for consistency, since they share the exact same `cache` + hand-rolled `onMounted(() => useDebounceFn(reload, 100)())` pattern as the list view.

That was reverted (see `39ef1c60` / `1b8796fe`). Those modals intentionally let a user close the builder without saving and reopen it later to resume unsaved edits — `auto: true` forces an immediate reload on every remount, which would silently discard that in-progress state. That's a regression for a "draft" editor, even though it's the correct behavior for a list view that should always reflect server truth. So this PR only changes `ViewControls.vue`; the layout modals are intentionally left on their original debounce-based reload.

PR #2492, opened 2026-07-17
fix(forms): description rendering — preserve line breaks + reliable auto-size

michellealva:
Render the form description with white-space: pre-wrap on the public page (.subtitle) and builder preview so newlines/spacing show as typed, and reliably auto-size the editor's description textarea on load/navigation (measure on an animation frame + retry until laid out) so a saved multi-line description no longer opens collapsed.


ps173 (approved):
LGTM!

PR #2505, opened 2026-07-18
fix: formatCurrency() ignores symbol_on_right — currency symbol always placed on left

### LINKED ISSUE #2402: formatCurrency() ignores symbol_on_right — currency symbol always placed on the left
## Description

`formatCurrency()` in `frontend/src/utils/numberFormat.js` hardcodes the currency symbol on the left side of the amount:

```javascript
return __(symbol) + ' ' + formatNumber(value, format, precision)
```

This follows the US convention (`$100`) but is incorrect for many currencies/locales where the symbol goes on the right (e.g. Polish Zloty: `100 zł`, Euro in many EU countries: `100 €`).

ERPNext's main Currency DocType has a `symbol_on_right` field that controls this, and Frappe's `format_currency()` Python function respects it. But FCRM's JavaScript `formatCurrency()` does not check this setting.

## Where it appears

The CRM Dashboard (`/crm/dashboard`) — all monetary values show the symbol on the left (e.g. `zł2.7M` instead of `2.7M zł`).

## Expected behavior

`formatCurrency()` should check the currency's `symbol_on_right` property (available via `frappe.boot.sysdefaults` or a `frappe.xcall` to get currency info) and place the symbol accordingly.

## Version

CRM 1.76.0, Frappe v16.25.0

georgekimaninjoroge:
Fixes #2402
formatCurrency() in numberFormat.js always placed the currency symbol on the left. For EUR that means € 400,000 instead of 400,000 €.
The fix adds getCurrencySymbolOnRight() which calls /api/resource/Currency/{code} to read the symbol_on_right field from the Currency DocType, caches the result at module level, and switches symbol placement in formatCurrency() accordingly.
One thing worth flagging: the lookup is async, so the first render still shows the symbol on the left. After the fetch resolves, the cache is populated and subsequent renders are correct. Happy to explore a different approach if there's a preferred pattern for preloading currency metadata in this app.
Before: € 400,000.00
After: 400,000.00 €
bug


shahzeelahmed:
@georgekimaninjoroge can you address the review?

georgekimaninjoroge:
@shahzeelahmed  fixed, should be okay now, just need review

frontend/src/main.js
```diff
@@ -51,25 +50,27 @@ for (let key in globalComponents) {
...
 app.config.globalProperties.$dialog = createDialog
 
 let socket
+async function mountApp() {
+  await preloadCurrencySymbolPlacement()
+  socket = initSocket()
+  app.config.globalProperties.$socket = socket
+  app.mount('#app')
+}
+
```
shahzeelahmed: page load will be blocked because this request. ideally no changes should be there in main.js just for currency symbol placement

frontend/src/components/SidePanelLayout.vue
```diff
@@ -546,14 +541,6 @@ function parsedField(field) {
...
 const attrs = instance?.vnode?.props ?? {}
 
-function isExternalUrl(value) {
-  return typeof value === 'string' && /^https?:\/\//i.test(value.trim())
-}
-
-function openExternalUrl(value) {
-  window.open(value.trim(), '_blank', 'noopener,noreferrer')
-}
-
```
shahzeelahmed, line 578: this is an ui regression and unrelated to the scope of this pr

georgekimaninjoroge: @shahzeelahmed done & done

ps173:
Hey can you please revert the removal of comments that you did. Some comments are required. Don't remove comments. Please attach a screenshot as well for this

PR #2507, opened 2026-07-19
fix: filter call logs by receiver/caller email instead of full name

### LINKED ISSUE #2145: Filter Issue: Call Logs
## Describe the bug

In the Call Logs list, clicking on a Receiver Name applies an Equals (`=`) filter instead of a partial (`LIKE`) filter. Because of this, no matching records are shown even when related data exists.

## To Reproduce

1. Go to `Call Logs`
2. Click on any `Receiver Name`
3. Observe the filtered results

## Current Behavior

The filter is being applied using an Equals match (`=`), which returns no results.

## Expected Behavior

The filter should use a partial match (`LIKE`) query  with (` % % `)for receiver fields so matching records are displayed correctly.

## Screenshot

Image

## Desktop

- OS: Windows 11
- Browser: Chrome

asimdelvi:
## Summary

Filtering Call Logs by clicking a "Caller" or "Receiver" cell was silently returning no results.

`caller` and `receiver` are `Link` fields to `User`, whose primar  key (`name`) is the user's email — but their stored/filterable value must be the email, not the full name. The row item objects built for these two columns in `getCallLogDetail()` (frontend/src/utils/callLog.js) only included `label` (full name) and `image`, with no `name`. The generic click-to-filter handler in `ViewControls.applyFilter()` resolves the filter value as `item.name ?? item.label ?? item`, so in the absence of `item.name` it fell back to the full name and sent that as the filter value — which never matches the stored email.

Other Link-to-User columns (`deal_owner`, `lead_owner`, `_assign`) already attach the email as `name` alongside `label`, so `applyFilter` picks the right value for them. This PR applies the same pattern to `caller`/`receiver`:

Fixes #2145

### DIFF
--- a/frontend/src/utils/callLog.js
+++ b/frontend/src/utils/callLog.js
@@ -15,11 +15,13 @@ export function getCallLogDetail(row, log, columns = []) {
     }
   } else if (row === 'caller') {
     return {
+      name: log.caller,
       label: log._caller?.label,
       image: log._caller?.image,
     }
   } else if (row === 'receiver') {
     return {
+      name: log.receiver,
       label: log._receiver?.label,
       image: log._receiver?.image,
     }


asimdelvi:
@ps173, can you check this PR?

ps173 (approved):
LGTM

PR #2516, opened 2026-07-21
fix: lead/deal report 500s when a child-table column is selected

kaulith:
## Summary

Restricted, in-hierarchy users opening the CRM Lead report and selecting a field from **CRM Products** hit:

```text
OperationalError (1052, "Column 'name' in IN/ALL/ANY is ambiguous")
```

This happened because the permission condition was rendered as `name IN (...)` instead of qualifying the column.
The query worked until a child-table field introduced a `LEFT JOIN`, making `name` ambiguous.

## Fix

Render the permission condition with `with_namespace=True` so it qualifies its own columns. 
The same fix is applied to both Lead and Deal.

---

Ref: https://support.frappe.io/helpdesk/tickets/74023?view=VIEW-HD+Ticket-1252

### DIFF
--- a/crm/permissions/org_hierarchy.py
+++ b/crm/permissions/org_hierarchy.py
@@ -62,12 +62,12 @@ def _permission_query_conditions(user: str | None, doctype: str):
 
 def get_lead_permission_query_conditions(user=None):
 	cond = _permission_query_conditions(user, "CRM Lead")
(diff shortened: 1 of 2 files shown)


kaulith:
Back-port needed.

shahzeelahmed (approved):

PR #2530, opened 2026-07-23
fix: preserve sender identity on received email replies

### LINKED ISSUE #2480: default CRM email sent as recipient, not user
**Describe the bug**

In the Frappe CRM module, replying to an email has no default email user. If not explicitly chosen in the to field droplist, the email is sent from the RECIPIENT email address which is obviously not valid. A from header is added with the email address of the recipient (copy of to header), which is rejected (rightly so) by the SMTP server.

**To Reproduce**
1. Go to a lead or deal or whatever
2. view emails
3. Click on reply
4. Compose and email
5. Send email - sender copies frappe user name and RECIPIENT email address.
**Expected behavior**

The default sender should match the sender from the prior address or current Frappe CRM user. 

Note- when clicking "new email" the sender address is pre-populated with the default (and can be changed via the droplist). I would expect that when hitting reply, the behavior would be the same as above- the current user email would be selected, rather than nothing.

In no case would I expect the recipient's email address be used as the sender, even if the current user doesn't have a valid email address set up in Frappe. In that case, it should fail.

**Screenshots**
Not applicable, as would expose confidential client/user information (email addresses) - can provide to a developer privately if requested. This appears in the the UI and in the Frappe logs. 

**Desktop (please complete the following information):**

MacOS/Safari
Frappe Framework: v16.20.0
ERPNext: v16.21.1
Wiki: v2.0.0
Frappe Builder: v1.0.0-dev
Print Designer: v1.6.7
Frappe CRM: v1.73.0

mateotiedra:
## Summary

- expose the canonical `Communication.sent_or_received` value in Lead and Deal activity payloads using one batched lookup
- restore a communication's historical sender only for explicitly `Sent` messages
- preserve the editor's authorized/default sender for `Received`, missing, or unknown direction values

Closes #2480.

## Context

#1859 added the authorized From selector and also made replies unconditionally assign `email.sender` to the editor's From value. That is useful for a `Sent` communication, where `sender` is the user's historical sending identity, but unsafe for a `Received` communication, where `sender` is the external correspondent.

This change keeps the intended behavior for sent messages while making received and legacy/missing-direction messages fail safe. Recipient, Reply All CC handling, subject, quoted content, attachments, drafts, and New Email behavior are unchanged.

`activities.py` was already near the repository's 500-line module limit, so its existing attachment/translation helpers were moved unchanged into `activity_helpers.py` alongside the batched direction lookup.

## Validation

Compared on the untouched `upstream/develop` base and the patched tree:

- full pre-commit suite (`pre-commit run --all-files`)
- frontend ESLint and Oxlint
- frontend Vitest coverage: 129/129 tests passed
- frontend production build
- full CRM server suite with coverage on a fresh local site
- existing Playwright email test: 2/2 setup/email tests passed
- isolated browser matrix covering New Email; Reply and Reply All for Received and Sent communications; multiple and single authorized identities; and a missing direction value

The browser matrix used only local `example.com` fixtures and did not click Send. The existing Playwright test created a local Communication using the repository's dummy localhost account; no real SMTP integration or external delivery was exercised.

## Backport

The regression is also present on the stable branch. Please consider this for the normal `main-hotfix` backport flow after merge.

### DIFF
--- a/crm/api/activities.py
+++ b/crm/api/activities.py
@@ -1,12 +1,11 @@
 import json
 
 import frappe
-from bs4 import BeautifulSoup
 from frappe import _
 from frappe.desk.form.load import get_docinfo
 from frappe.query_builder import JoinType
-from frappe.translate import get_translated_doctypes
 
+from crm.api.activity_helpers import get_communication_directions, is_translatable, parse_attachment_log
 from crm.fcrm.doctype.crm_call_log.crm_call_log import parse_call_log
 
 
@@ -26,6 +25,7 @@ def get_deal_activities(name: str):
 
 	get_docinfo("", "CRM Deal", name)
 	docinfo = frappe.response["docinfo"]
+	communication_directions = get_communication_directions(docinfo.communications)
 	deal_meta = frappe.get_meta("CRM Deal")
 	deal_fields = {
 		field.fieldname: {"label": field.label, "options": field.options} for field in deal_meta.fields
@@ -147,6 +147,7 @@ def get_deal_activities(name: str):
 				"attachments": get_attachments("Communication", communication.name),
 				"read_by_recipient": communication.read_by_recipient,
 				"delivery_status": communication.delivery_status,
+				"sent_or_received": communication_directions.get(communication.name),
 			},
 			"is_lead": False,
 		}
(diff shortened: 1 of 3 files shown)


mateotiedra:
Confirmed during the isolated API/UI verification: `get_docinfo()` communication objects did not hydrate `sent_or_received`. The backing Communication rows had `Sent`/`Received`, but reading the field from the docinfo objects produced missing values; the activity response only returned the canonical directions after the batched lookup. The helper keeps this to one query for all communications rather than one query per row.

shahzeelahmed:
@mateotiedra please resolve conflicts

PR #2532, opened 2026-07-24
fix: guard filters before list params initialize

### LINKED ISSUE #2509: Quick filters throw TypeError: params is null and fail to apply unless main filter bar used first
Issue Statement:

Quick filters in the CRM list view (Leads, Deals, etc.) throw a JavaScript error and fail to apply unless a filter has already been set from the main filter bar first.

Issue Detailed Description:

When a quick filter (e.g. a Link-type field like Lead Owner) is applied from the quick filter row above the list, the frontend throws TypeError: can't access property "filters", d.value.params is null in ViewControls.vue. This happens because the underlying list resource's params object can be transiently null (e.g. right after switching views, kanban reload, or before the initial data fetch completes), and the quick filter handler (applyQuickFilter) reads list.value.params.filters without a null check. The same unguarded pattern also exists in the "filter from list cell" and "liked by me" filter handlers. Filtering via the main filter bar works around this because, once any filter has been applied that way, list.value.params has already been populated — after which quick filters also start working.

Steps to Replicate the issue:

1. Open any CRM list view (e.g. Leads or Deals) fresh, without applying any filter from the main filter bar.
2. Click into a quick filter field above the list (e.g. a Link field such as "Lead Owner" or "Assigned To").
3. Select/enter a value in that quick filter.
4. Open the browser console.
5. Current behavior:

The quick filter does not apply. The browser console shows:

> TypeError: can't access property "filters", d.value.params is null
>     at ViewControls.vue:836 (applyQuickFilter)
> The list is not filtered, and the error can recur on subsequent quick filter attempts until a filter is first applied from the main filter bar.

Expected behavior:

Selecting a value in a quick filter should apply the filter to the list immediately, with no console error, regardless of whether a filter has previously been set from the main filter bar.

Supporting Screenshots/documents/videos/dashboard links:

 
Image

thangle-osg:
## Summary

Prevent quick-filter actions from throwing when the list resource has not
initialized `params` yet.

Fixes #2509.

This replaces #2531, which Mergify closed because it targeted the stable
`main` branch. This branch is based on `develop`.

## Root cause

`ViewControls.vue` directly spread `list.value.params.filters` in three paths:

- Applying a quick filter
- Filtering by a list cell value
- Toggling the "liked by me" filter

The Vue component can render and accept input while the Frappe UI list resource
is still initializing. During that window, the reactive resource exists but
`list.value.params` can be null. JavaScript therefore throws while evaluating
`params.filters`, before the handler reaches `updateFilter()`.

Using the main filter dialog first can hide the problem because that path
initializes params as a side effect.

## Fix

Add one initialization-boundary helper and use it in all three handlers:

```js
function getCurrentFilters() {
  return { ...(list.value.params?.filters ?? getParams().filters) }
}
```

This approach:

- Keeps the asynchronous initialization rule in one place.
- Clones filters before a handler modifies them.
- Uses initialized resource filters when available.
- Reconstructs filters from the active view while resource params are absent,
  preserving existing view state rather than falling back to an empty object.
- Leaves the existing `updateFilter()` and resource reload workflow unchanged.

## Validation

- `yarn test:run`: 7 test files passed, 129 tests passed.
- `yarn prettier --check src/components/ViewControls.vue`: passed.
- A downstream production build with the same component change passed,
  including PWA generation.

### DIFF
--- a/frontend/src/components/ViewControls.vue
+++ b/frontend/src/components/ViewControls.vue
@@ -520,6 +520,11 @@ function getParams() {
   }
 }
 
+/** Return cloned active filters even before resource params initialize. */
+function getCurrentFilters() {
+  return { ...(list.value.params?.filters ?? getParams().filters) }
+}
+
 list.value = createResource({
   url: 'crm.api.doc.get_data',
   params: getParams(),
@@ -864,7 +869,7 @@ function setupNewQuickFilters(filters) {
 }
 
 function applyQuickFilter(filter, value) {
-  let filters = { ...list.value.params.filters }
+  let filters = getCurrentFilters()
   let field = filter.fieldname
   if (value) {
     if (
@@ -1314,7 +1319,7 @@ function applyFilter({ event, idx, column, item, firstColumn }) {
   event.stopPropagation()
   event.preventDefault()
 
-  let filters = { ...list.value.params.filters }
+  let filters = getCurrentFilters()
 
   let value = item.name ?? item.label ?? item
 
@@ -1339,7 +1344,7 @@ function applyFilter({ event, idx, column, item, firstColumn }) {
 }
 
 function applyLikeFilter() {
-  let filters = { ...list.value.params.filters }
+  let filters = getCurrentFilters()
   if (!filters._liked_by) {
     filters['_liked_by'] = ['LIKE', '%@me%']
   } else {


shahzeelahmed:
this was fixed in a different PR

AL2018:
This PR was closed as fixed by #2548, but #2548's fix does not work.

#2548 adds list.value.params = getParams() after the resource is created. list is a defineModel, and its getter has not caught up with the setter at that point in setup — so the params land on the {} default rather than on the resource. Instrumented on a running build: modelGetter returns the resource? false | url: undefined, and params is still null three seconds later.

Verified still broken on v1.80.0 and v1.81.2. The develop branch carries the same shape.

The approach in this PR — falling back on read rather than writing through the model — avoids that problem entirely, which is why I think it was closed in error.

One addition worth making if it is reopened: exportRows() reads params three times, including order_by. A fallback there matters, because an undefined order_by interpolates into the export URL as order_by=undefined rather than failing loudly.

Root cause is frappe/crm@96d100d0 (15 July), which added auto: true and removed the onMounted reload that was the only thing assigning params. Context in #2708.

PR #2539, opened 2026-07-25
build: add production docker compose stack

OXRED9:
The existing docker/docker-compose.yml is a development bench: it bind-mounts init.sh from the host and runs `bench init` at container start, so it cannot survive a deploy platform's startup health check. It has no host mount to read its entrypoint from, and binds no port until a 10-30 minute build completes.

Add docker/production/ as a deployable alternative, adapted from upstream frappe_docker's pwd.yml:

- Runs the prebuilt ghcr.io/frappe/crm image instead of building at startup.
- Publishes a configurable CRM_HTTP_PORT (default 8052) onto nginx's 8080.
- Pins FRAPPE_SITE_NAME_HEADER to the site name so health probes hitting a bare IP resolve to the site instead of erroring with "Site does not exist".
- Uses `restart:` rather than `deploy.restart_policy`, which plain Compose ignores outside Swarm.
- Adds a frontend healthcheck against /api/method/ping with a start_period wide enough to cover first-boot site creation.

Also gitignore .env so filled-in deployment secrets are not committed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ e2e/.auth
 test-results
 playwright-report
 package-lock.json
+.env
 
 .vscode
 .github/instructions
--- a/docker/production/.env.example
+++ b/docker/production/.env.example
@@ -1,0 +1,21 @@
+# Copy to .env next to docker-compose.yml and fill in real values.
+# Do not commit the filled-in .env.
+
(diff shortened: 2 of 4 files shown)


OXRED9:
ok

PR #2557, opened 2026-07-28
feat(views): lucide icon picker for views

michellealva:
Replaces the emoji-only icon picker in the view modal with frappe-ui's **lucide `IconPicker`** (searchable, ~1000 icons) — the same approach the latest Helpdesk uses.

### Changes
- **`main.js`** — register frappe-ui's `spritePlugin` so the lucide SVG sprite is injected into the DOM. The `IconPicker` and lucide `Icon`s read icon names from it (without this the picker shows *"No icons available"*).
- **`Modals/ViewModal.vue`** — use `IconPicker` from `frappe-ui/icons`. Keeps a **legacy-emoji fallback**: a view that stored an emoji shows the emoji beside the picker and preserves it until a lucide icon is picked.
- **`Icon.vue`** — render string icon names from the lucide sprite (no fixed width/height, so size classes still control sizing, matching the old FeatherIcon). Emoji and component branches are unchanged, so existing emoji icons keep rendering.

### Migration
Backward-compatible, **per-view opt-in** — no data patch. Existing emoji icons keep rendering and migrate individually only when a view is edited and a lucide icon chosen. Standard views (List/Kanban/Group By) use component icons and are unaffected.

### Screenshots

image

### DIFF
--- a/frontend/src/components/Icon.vue
+++ b/frontend/src/components/Icon.vue
@@ -2,11 +2,22 @@
   <div v-if="isEmoji(icon)" v-bind="$attrs">
     {{ icon }}
   </div>
-  <FeatherIcon
+  <!-- lucide icon from the sprite the IconPicker reads (lucide is a superset of
+       feather, so legacy names still resolve). No width/height attrs so size
+       classes like `h-4` control it, matching the old FeatherIcon behaviour. -->
+  <svg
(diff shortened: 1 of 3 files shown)


ps173 (approved):
LGTM!

PR #2563, opened 2026-07-28
fix(telephony): play call recordings reliably

### LINKED ISSUE #2550: Cant play the recording in call logs
Audio thing broken

Image

michellealva:
Fixes #2550.

The call-log recording player was broken in two ways: recordings that *did* exist showed a **0:00 duration** with a ~2-3s delay before playing (and couldn't seek), and calls whose `recording_url` no longer resolves to a real recording showed a **dead empty player**.

### Changes
- **`get_recording_url` (`crm/integrations/api.py`)** — stream the provider's response instead of buffering the whole file, and forward the browser's `Range` header plus the provider's `Content-Length`/`Content-Range` (with `Accept-Ranges`). The HTML `<audio>` element can now read the duration **immediately** and seek, instead of waiting for the entire recording to download server-side first. (Verified: a range request returns `206` with `Content-Range` in ~0.7s vs the whole file buffering for 2-3s.)
- **`CallLogDetailModal.vue`** — when the recording fails to load (`recording_url` set but no real/expired recording), show **"Recording not available"** instead of a dead 0:00 player. Resets per call log.

### Testing
Manual (needs a telephony provider + a real recording, so no automated test): verified with a Range-capable audio source that the endpoint returns `206 Partial Content` with correct `Content-Range`/`Content-Length` quickly, and that a non-resolving `recording_url` renders the fallback.

### DIFF
--- a/crm/fcrm/doctype/crm_call_log/test_crm_call_log.py
+++ b/crm/fcrm/doctype/crm_call_log/test_crm_call_log.py
@@ -1,6 +1,8 @@
 # Copyright (c) 2023, Frappe Technologies Pvt. Ltd. and Contributors
 # See license.txt
 
+from unittest.mock import MagicMock, patch
+
 import frappe
 from frappe.tests import IntegrationTestCase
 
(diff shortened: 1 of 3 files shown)


ps173 (approved):
LGTM

michellealva:
fixed in 3aa5d89a. The name watch didn't fire when the *same* call log was reopened, so `recordingError` stayed stale and hid a now-playable recording. Added a watch on the modal's `show` that resets `recordingError` whenever it opens, so reopening the same log re-attempts the load.

michellealva:
@greptile review

PR #2581, opened 2026-07-29
fix(assign_to): use public add() API instead of removed private _add()

### LINKED ISSUE #2413: bench migrate fails on CRM Lead: ImportError: cannot import name '_add' from 'frappe.desk.form.assign_to'
## Environment
- Frappe Framework: v16.20.0 (branch `version-16`)
- Frappe CRM: v1.76.0 (branch `main`)
- Python: 3.14
- Hosting: Frappe Cloud

## Summary
Running `bench migrate` on a site with both Frappe Framework (current `version-16`) and CRM installed fails while syncing DocTypes for the `crm` app. `crm/crm/fcrm/doctype/crm_lead/crm_lead.py` imports a private symbol `_add` from `frappe.desk.form.assign_to`, but that function is named `add` (no underscore, `@frappe.whitelist()`-decorated) in the current Frappe Framework -- there is no `_add` in `frappe/desk/form/assign_to.py` at all. This breaks module import for the `CRM Lead` DocType, which aborts the entire site migration (not just the CRM Lead doctype), since `frappe.model.sync.sync_all()` fails hard on the first import error it hits.

## Steps to reproduce
1. Install Frappe Framework (current `version-16`, e.g. v16.20.0) and CRM (v1.76.0, `main`) on the same site.
2. Run `bench migrate` (or trigger a Frappe Cloud site update).
3. Migration proceeds through `frappe`, `erpnext`, `hrms`, `india_compliance`, `print_designer`, `payments` DocTypes successfully, then fails partway through `crm` DocTypes (~28% through, on `CRM Lead`).

## Expected behavior
`bench migrate` completes successfully.

## Actual behavior
Migration aborts with:
```
ImportError: cannot import name '_add' from 'frappe.desk.form.assign_to' (.../apps/frappe/frappe/desk/form/assign_to.py)
```
Full traceback (relevant part):
```
File ".../apps/crm/crm/fcrm/doctype/crm_lead/crm_lead.py", line 8, in <module>
    from frappe.desk.form.assign_to import _add as assign
ImportError: cannot import name '_add' from 'frappe.desk.form.assign_to'

...

File ".../apps/frappe/frappe/modules/utils.py", line 308, in load_doctype_module
    raise ImportError(msg) from e
ImportError: Module import failed for CRM Lead, the DocType you're trying to open might be deleted.
Error: cannot import name '_add' from 'frappe.desk.form.assign_to' (.../apps/frappe/frappe/desk/form/assign_to.py)
```
Because `sync_all()` fails outright on this import error, the site's migrate step never proceeds past `crm`, so no further schema updates for the site apply and (on Frappe Cloud specifically) the pending deploy is rolled back / recovered rather than going live.

## Root cause
In the current Frappe Framework, `frappe/desk/form/assign_to.py` only defines `add(args=None, *, ignore_permissions=False)` -- there's no `_add` alias/private function. `crm_lead.py` still does:
```python
from frappe.desk.form.assign_to import _add as assign
```
which suggests `_add` was renamed to `add` upstream in Frappe Framework at some point, and this particular CRM import was never updated to match.

## Suggested fix
In `crm/crm/fcrm/doctype/crm_lead/crm_lead.py`, change:
```python
from frappe.desk.form.assign_to import _add as assign
```
to:
```python
from frappe.desk.form.assign_to import add as assign
```
(and audit the rest of the CRM codebase for other `_add`/similar private-import references to the same module, in case this isn't the only call site.)

## Impact
Blocks `bench migrate` -- and therefore any Frappe Cloud site update/deploy that touches migration -- on any site running CRM alongside a current `version-16` Frappe Framework, regardless of what else is being deployed at the time.

tareko:
**Note: This PR is AI-assisted**

## Summary

`CRM Lead` and `CRM Deal` import the **private** `_add()` from `frappe.desk.form.assign_to`. That symbol was removed in some Frappe 16.x patch releases (it is present on `version-16` HEAD, but absent in others — e.g. **16.5.0**, shipped inside `frappe/erpnext:v16.4.1`). On those Frappe versions, `bench install-app crm` / `bench migrate` abort while syncing the doctypes:

```
ImportError: cannot import name '_add' from 'frappe.desk.form.assign_to'
```

This switches Lead/Deal to Frappe's **public, stable** `add()` — the same API `CRM Task` already uses (`crm/fcrm/doctype/crm_task/crm_task.py`).

> Retargeted to `develop` from #2580 per review.

## Why this is correct

- `add(args)` is the `@frappe.whitelist()`-decorated public entry point; `_add()` is the internal worker whose existence varies across Frappe versions.
- `add()` does not accept `ignore_permissions=`, so the kwarg is dropped from the two call sites. This is safe: `assign_agent()` is invoked by a user who already holds permission on the Lead/Deal (they are editing/creating it), so the permission check inside `add()` passes — exactly the path `CRM Task.assign_to()` already relies on.
- Behaviour is unchanged for the common case; the only difference is that an explicit permission check now runs (consistent with `CRM Task`).

## Prior art

#2413 reported this and #2380 was merged as the fix, but **the offending `import _add` was not actually removed** — it is still present on both `main` and `develop`:

- `crm/fcrm/doctype/crm_lead/crm_lead.py:8`
- `crm/fcrm/doctype/crm_deal/crm_deal.py:6`

This PR removes it.

## Test

Validated against `frappe/erpnext:v16.4.1` (which bundles **Frappe Framework 16.5.0**, where `_add` is absent). On a throwaway site, without the fix `bench --site <site> install-app crm` fails with the `ImportError` above; **with this fix it completes cleanly**:

```
Installing crm...
Updating DocTypes for crm: 100%
* Installing Custom Fields in Email Template / Email Account / Web Form / Assignment Rule
Creating Workspace Sidebars
Creating Sidebar Items for Frappe CRM
Updating Dashboard for crm      # exit 0
```

Result: 39 FCRM doctypes created, `crm` registered in `Installed Application`. No `ImportError`.

Fixes #2413.

### DIFF
--- a/crm/fcrm/doctype/crm_deal/crm_deal.py
+++ b/crm/fcrm/doctype/crm_deal/crm_deal.py
@@ -3,7 +3,7 @@
 
 import frappe
 from frappe import _
-from frappe.desk.form.assign_to import _add as assign
+from frappe.desk.form.assign_to import add as assign
 from frappe.model.document import Document
 
 from crm.api.exchange_rate import get_exchange_rate
@@ -172,7 +172,7 @@ class CRMDeal(Document):
 					# the agent is already set as an assignee
 					return
 
-		assign({"assign_to": [agent], "doctype": "CRM Deal", "name": self.name}, ignore_permissions=True)
+		assign({"assign_to": [agent], "doctype": "CRM Deal", "name": self.name})
 
 	def share_with_agent(self, agent):
 		if not agent:
--- a/crm/fcrm/doctype/crm_lead/crm_lead.py
+++ b/crm/fcrm/doctype/crm_lead/crm_lead.py
@@ -5,7 +5,7 @@ import json
 
 import frappe
 from frappe import _
-from frappe.desk.form.assign_to import _add as assign
+from frappe.desk.form.assign_to import add as assign
 from frappe.model.document import Document
 from frappe.utils import validate_email_address
 
@@ -177,7 +177,7 @@ class CRMLead(Document):
 					# the agent is already set as an assignee
 					return
 
-		assign({"assign_to": [agent], "doctype": "CRM Lead", "name": self.name}, ignore_permissions=True)
+		assign({"assign_to": [agent], "doctype": "CRM Lead", "name": self.name})
 
 	def share_with_agent(self, agent):
 		if not agent:


tareko:
Bump as I believe this issue is still live.

PR #2597, opened 2026-07-30
fix: add phone validation to from/to number fields in crm call log

### LINKED ISSUE #2566: Call Log accepts any string in "From" and "To" phone number fields — no validation
**Describe the bug**
When creating a Call Log manually, the "From" and "To" fields accept any string (letters, symbols, junk text) with no validation. Invalid entries save successfully, which pollutes call records and breaks any downstream matching of call logs to leads/contacts by phone number.

Shindhu-Ramaswamy:
Fixes #2566

**What changed**
Added `"options": "Phone"` to the `From Number` and `To Number` fields
in the CRM Call Log doctype. This is Frappe's built-in "Phone" data-field
option, which:
- Rejects non-phone-format text on save (server-side validation)
- Flags invalid input in the desk form UI as you type (client-side validation)

Guarded the Twilio `voice` webhook 
   against a fallout of (1): when an agent places an outgoing call from
   the browser (Twilio Client) but has no phone number mapped in their
   CRM Telephony Agent record, Twilio's raw request falls back to a
   client identity string (e.g. `client:agent_john`) instead of a real
   number. Previously this string would flow into the Call Log's `from`
   field; with the new Phone validation it would instead fail with an
   unhandled error back to Twilio. The webhook now checks for this case
   upfront and responds with a clear voice message
   ("Your account is not configured with a phone number...") instead of
   attempting the call, so no invalid Call Log is ever created.
   (Exotel's outgoing-call path already guards this same scenario via
   `frappe.throw` — this brings Twilio in line with that.)

**Tested**
- Invalid input ("abc123") → validation error on save ✓
- Valid formats (+91 98765 43210, 022-1234-5678) → saved ✓

### DIFF
--- a/crm/fcrm/doctype/crm_call_log/crm_call_log.json
+++ b/crm/fcrm/doctype/crm_call_log/crm_call_log.json
@@ -43,6 +43,7 @@
    "in_list_view": 1,
    "in_standard_filter": 1,
    "label": "From Number",
+   "options": "Phone",
    "reqd": 1
   },
   {
@@ -82,6 +83,7 @@
    "in_list_view": 1,
    "in_standard_filter": 1,
    "label": "To Number",
+   "options": "Phone",
    "reqd": 1
   },
   {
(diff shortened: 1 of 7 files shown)


ps173:
@Shindhu-Ramaswamy Add validation for the frontend form as well. Also the source of the invalid numbers is CRM Telephony Agent. Fix the validation in that as well.

ps173 (approved):

PR #2652, opened 2026-08-10
fix: add Edit option to Notes three-dot menu

### LINKED ISSUE #2605: Notes: "Edit" option missing from three-dot menu (inconsistent with Tasks)
In the Notes tab of a Lead/Deal, the three-dot menu on a note shows only
**Delete**. Notes are editable — clicking anywhere on the note card opens
the editor — but there is no explicit Edit option in the menu.

Shindhu-Ramaswamy:
Fixes #2605

Added the explicit edit option in Notes

### DIFF
--- a/frontend/src/components/Activities/Activities.vue
+++ b/frontend/src/components/Activities/Activities.vue
@@ -45,7 +45,11 @@
           :key="note.name"
           @click="modalRef.showNote(note)"
         >
-          <NoteArea v-model="all_activities" :note="note" />
+          <NoteArea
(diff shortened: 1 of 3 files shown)


ps173 (approved):
LGTM!

note: maybe we can remove the click event from the note as well later. This works for now

Shindhu-Ramaswamy:
[Screencast from 2026-08-10 20-13-56.webm](https://github.com/user-attachments/assets/e0ff62e5-08c4-45c8-8ae6-ed4fe0991d29)

Issue #2653, opened 2026-08-11
CRM frontend calls missing Frappe Pulse `boot_config` method and breaks logged-in navigation


yoc-san:
### Summary

After updating to the latest available versions, Frappe CRM still calls:

```js
frappe.utils.telemetry.pulse.client.boot_config
```
but the current Frappe v15 Pulse telemetry client does not expose a boot_config method.
This causes logged-in navigation to fail with a server-side ValidationError.
Environment

- Frappe: version-15, latest upstream at the time of testing
- observed commit/tag: 15.117.0
- Frappe CRM: latest upstream main at the time of testing
- observed tag: v1.81.1
- Logged-in user session
- Frappe site managed through Docker/frappe-manager

```
Error
AttributeError: module 'frappe.utils.telemetry.pulse.client' has no attribute 'boot_config'
frappe.exceptions.ValidationError: Failed to get method for command frappe.utils.telemetry.pulse.client.boot_config with module 'frappe.utils.telemetry.pulse.client' has no attribute 'boot_config'
```
Where the call happens

In CRM:
`// apps/crm/frontend/src/router.js`

```
const { enabled } =
  (await call('frappe.utils.telemetry.pulse.client.boot_config')) || {}
return !!enabled
```

What Frappe currently exposes
In Frappe:
`apps/frappe/frappe/utils/telemetry/pulse/client.py`

The module exposes whitelisted methods such as:

`is_enabled`
`capture`
`bulk_capture`
`get_debug_info`

but does not expose:
`boot_config`

### Expected behavior
Telemetry should never block logged-in navigation.
If Pulse boot config is unavailable, CRM should gracefully degrade telemetry to disabled instead of throwing an exception.
### Actual behavior
The CRM frontend calls a missing whitelisted method. Frappe raises a server-side error, and the logged-in navigation flow breaks.
### Related observation
Helpdesk appears to handle this case defensively:
```
try:
    from frappe.utils.telemetry.pulse.client import boot_config

    return boot_config()
except Exception:
    return {"enabled": False}
```
This suggests telemetry boot should degrade gracefully when unavailable.
### Suggested fix
In CRM frontend, wrap the telemetry boot call defensively and fall back to is_enabled or disabled telemetry.
Example:
```
async function shouldCapturePersona() {
  if (localStorage.getItem(PERSONA_DONE_KEY)) return false

  const captured = await call('frappe.client.get_single_value', {
    doctype: 'FCRM Settings',
    field: 'persona_captured',
  })

[...]

PR #2661, opened 2026-08-13
fix(router): stop the persona-check crashing server-side on Frappe v15

suparikoli:
## What's actually still wrong here, vs. #2653 as filed

#2653 reports that `boot_config` is missing on Frappe v15 and that this breaks logged-in navigation. Before writing anything I checked both halves against real source, and they don't fully hold up as stated — worth being precise about, since it changes what this PR is:

- **`boot_config` missing on v15 is real.** `frappe/frappe`'s `version-15` branch has no `boot_config` in `pulse/client.py` at all (only `is_enabled`, `capture`, `bulk_capture`, …). `develop` has it, whitelisted.
- **But `is_enabled` — the endpoint CRM called before #142e7a07 — has the opposite problem: it's no longer whitelisted on `develop`.** So there is currently no single endpoint that's callable from both an older stable Frappe site and a newer one. Whichever one CRM calls, some real deployment breaks.
- **"Breaks logged-in navigation" isn't reproducible on current `develop`.** #ffcc3032 wrapped the whole `shouldCapturePersona()` call site (both call sites in the router guard) in try/catch back in July, for an unrelated reason, before the `boot_config` switch landed. That catch is still there, so the promise rejection this issue describes is already caught and doesn't propagate to break routing.

What's real and still unfixed: **every navigation on a v15 site throws a server-side `ValidationError`/`AttributeError` for a method that will never resolve, gets silently swallowed by the outer catch, and nothing about it is visible anywhere** — not the console, not a log line, nothing. The onboarding wizard just silently never fires on any v15 install, and no one can tell why. That's what this PR fixes.

## Fix

Move the try/catch to wrap the `boot_config` call specifically, rather than relying on the outer router guard's broader catch to mask it. Default `enabled` to `false` on failure — the check gates nothing but an onboarding wizard, so "can't tell if telemetry is enabled" safely means "don't show it," matching what the outer catch was already doing by accident, just now intentional and scoped to the one call that can actually fail.

Not attempting the harder, more general problem (Frappe having no client-callable endpoint that works across both branches) — that's a decision for Frappe core, not something to work around further in CRM's router.

## Testing

`shouldCapturePersona` wasn't exported or covered by any test. Exported it and added `frontend/tests/unit/shouldCapturePersona.test.js` — mocks `frappe-ui`'s `call`, no server needed. **Actually run**, not just reasoned about:

```
✓ tests/unit/shouldCapturePersona.test.js (5)
  ✓ returns false without calling the server when already captured locally
  ✓ returns false when persona_captured is already set server-side
  ✓ returns true when boot_config resolves enabled
  ✓ returns false when boot_config resolves disabled
  ✓ returns false, without throwing, when boot_config is not available
```

The last case is the regression test: rejects the `boot_config` call the way a v15 site's server actually does, and asserts the function resolves `false` rather than the rejection propagating.

### DIFF
--- a/frontend/src/router.js
+++ b/frontend/src/router.js
@@ -25,7 +25,7 @@ export async function shouldCapturePersona() {
   // "telemetry is not enabled": this call gates nothing but the onboarding
   // wizard, so the safe default is to skip it, not to let a missing
   // whitelisted method surface as a server-side error on every navigation.
-  let enabled = false
+  let enabled
   try {
     const config = await call('frappe.utils.telemetry.pulse.client.boot_config')
     enabled = config?.enabled
--- a/frontend/src/router.js
+++ b/frontend/src/router.js
@@ -7,7 +7,7 @@ import { viewsStore } from '@/stores/views'
 let personaChecked = false
 export const PERSONA_DONE_KEY = 'crm_persona_captured'
 
-async function shouldCapturePersona() {
+export async function shouldCapturePersona() {
   // Client-side flag guards against re-prompting if the server persist failed.
   if (localStorage.getItem(PERSONA_DONE_KEY)) return false
   const captured = await call('frappe.client.get_single_value', {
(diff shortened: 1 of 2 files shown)


frontend/src/router.js
```diff
@@ -16,8 +16,22 @@ async function shouldCapturePersona() {
...
+  // version to break, treat "can't tell if telemetry is enabled" as
+  // "telemetry is not enabled": this call gates nothing but the onboarding
+  // wizard, so the safe default is to skip it, not to let a missing
+  // whitelisted method surface as a server-side error on every navigation.
+  let enabled
+  try {
+    const config = await call('frappe.utils.telemetry.pulse.client.boot_config')
+    enabled = config?.enabled
+  } catch {
+    enabled = false
```
shahzeelahmed, line 33: **[P1] Preserve telemetry onboarding on Frappe v15**

Catching the missing `boot_config` call and returning `false` makes the failure quieter, but it also permanently disables persona onboarding on every telemetry-enabled v15 site. v15 exposes `is_enabled()`, while newer Frappe keeps the same function internally but replaces its public endpoint with `boot_config()`. CRM should bridge that API difference instead of treating v15 as telemetry-disabled.

A clean cross-version solution is to add `telemetry_enabled: is_pulse_enabled()` to the existing CRM `get_boot()` response and read `window.telemetry_enabled` here. `frappe.utils.telemetry.is_pulse_enabled` is exported on both v15 and newer Frappe, and the CRM boot keys are already exposed on `window`. This avoids version detection, failed endpoint probes, and an extra request. The regression test should then verify enabled and disabled behavior rather than enshrine the rejected-call fallback.

PR #2667, opened 2026-08-13
feat: next follow up date on leads & deals with reminder notification

### LINKED ISSUE #2588: Add Follow-Up Date field on Leads & Deals with automatic reminder notification to the assigned user
**Is your feature request related to a problem? Please describe.**

There is currently no first-class way to answer *"when am I next supposed to touch this lead/deal, and who is responsible?"*

Following up is the core loop of a sales process, but today a rep has to improvise a place to record the next touch date:

- Create a **Task** with a due date — works, but a task is a separate record, so the follow-up date isn't visible on the Lead/Deal list, can't be sorted/filtered as a column, and there is no single "next follow-up" per record.
- Put it in a **comment** or the `next_step` text field — invisible to sorting, filtering, and any automation.
- Add a **custom field** — this is what people end up doing (the docs even use `{ fieldname: 'follow_up', fieldtype: 'Date', label: 'Follow Up' }` as the example custom field in `.pi/feats/form-scripting/form-dialog.md`), but a custom field is inert: nothing reminds anyone when the date arrives.

The result is that follow-ups get missed. The date either lives somewhere the system can't act on, or it lives on a task disconnected from the deal's own timeline. There is no built-in follow-up field on **CRM Lead** or **CRM Deal**, and no reminder tied to one.

**Describe the solution you'd like**

Two connected pieces:

**1. A `next_follow_up` datetime field on CRM Lead and CRM Deal**

- Type `Datetime` (a time matters — "call at 3pm", not just "sometime Tuesday").
- Available as a list-view column, sortable and filterable, so a rep can pull up "my follow-ups due today" and a manager can spot overdue ones.
- Surfaced in the side panel / detail view like other key fields.

**2. An automatic reminder notification to the assigned user when that date arrives**

- When the follow-up datetime is reached, notify the user(s) assigned to the record (`_assign`).
- Delivered as an **in-app notification** (`CRM Notification`), with **email** as a configurable option.
- Ideally a configurable lead time ("remind me 30 minutes / 1 day before"), with a sensible default.
- The reminder should follow the record: if the follow-up date is changed the reminder reschedules, and if the date is cleared or the record is closed/won/lost the pending reminder is dropped so it doesn't fire on a dead deal.
- If nobody is assigned, fall back to the owner rather than silently sending nothing.

**Why this belongs in core rather than as a custom field:** the field alone is easy to add, but the *reminder* is the part that actually prevents missed follow-ups, and that needs scheduler + notification wiring that users can't reasonably bolt on themselves.

**Describe alternatives you've considered**

- **Tasks with due dates** — the closest existing option, and #1798 covered task due-date reminders. But a task is a separate object; it doesn't give the Lead/Deal itself a sortable "next touch" date, and reps think in terms of "when do I next contact this deal", not "which task record is attached".
- **Custom field + Notification / Server Script** — possible, but every deployment has to reinvent the scheduling, the reschedule-on-change behavior, and the assignee resolution. It's also easy to get wrong (duplicate or orphaned reminders).
- **Relying on the existing Event/calendar reminders** — Events already have reminders, but a follow-up isn't naturally a calendar event, and creating one per follow-up is friction on the exact action that needs to be fast.

**Additional context**

Most of the machinery for this already exists in the codebase, so this looks more like wiring than net-new infrastructure:

- **`CRM Notification` + `notify_user()`** in `crm/fcrm/doctype/crm_notification/crm_notification.py` — already used by `crm/api/comment.py`, `crm/api/todo.py`, and `crm/api/whatsapp.py`.
- **A working offset-reminder precedent** in `crm/api/event.py`: `trigger_offset_event_notifications()` plus `_process_event_notifications_by_interval()`, `_send_system_notification()`, and `_send_email_notification()` already implement "notify N units before a datetime", registered via `scheduler_events` in `crm/hooks.py` (`all` / `hourly` / `daily` / `weekly`). A follow-up reminder could reuse this same shape.
- **Assignment** is already `_assign` on both `crm/fcrm/doctype/crm_lead/crm_lead.py` and `crm/fcrm/doctype/crm_deal/crm_deal.py`, so resolving "the assigned person" is straightforward.

**Related issues:** #1866 requests a general scheduled-reminder system for CRM activities, and #1798 (closed) covered task due-date reminders. This issue is deliberately narrower: a dedicated **follow-up date field on Lead/Deal** plus a reminder bound to it. Happy to have it folded into #1866 as a concrete sub-case if maintainers prefer a single unified reminder framework.

chinmaybhatk:
Closes #2588

## Problem

There is no first-class way to answer *"when am I next supposed to touch this lead/deal, and who is responsible?"*. Today a rep has to improvise: a separate Task record (invisible on the Lead/Deal list, no single "next touch"), a comment or `next_step` (invisible to sorting, filtering and automation), or a custom field — which is what people end up doing, but a custom field is inert, so nothing reminds anyone when the date arrives.

## What this adds

**1. A `next_follow_up` Datetime on CRM Lead and CRM Deal**

- `Datetime`, because the time matters ("call at 3pm", not "sometime Tuesday")
- Indexed, and added to `default_list_data` so it is a sortable/filterable list column — a rep can pull up "my follow-ups due today", a manager can spot overdue ones
- Added to the default Lead/Deal Side Panel layouts

**2. A reminder when it comes due**

`crm.api.follow_up.trigger_follow_up_reminders` runs on the `all` scheduler event and notifies the users responsible for the record:

- In-app `CRM Notification` (new `Follow Up` type), with email as a configurable option
- Recipients are `_assign`, falling back to `lead_owner`/`deal_owner` and then the record owner, so a reminder is never silently dropped when nobody is assigned
- Configurable lead time — *n* minutes/hours/days, default 30 minutes — alongside an enable switch and the email option, in **Settings → General**

This reuses the existing machinery the issue pointed at: `notify_user()` from `CRM Notification`, and the same "notify N units before a datetime" shape as `crm/api/event.py`.

## Reminder lifecycle

Rather than a queue of pending reminder rows, state is a single hidden `follow_up_reminder_sent` checkbox on the record, cleared in `validate` whenever `next_follow_up` changes. That makes the three lifecycle rules from the issue nearly free:

| Case | Behaviour |
| --- | --- |
| Rescheduled | flag clears on save, the new time gets its own reminder |
| Cleared | no date, so the record never matches the query |
| Won / Lost / converted lead | filtered out, so a dead deal never fires |

Overdue follow-ups still notify (the scheduler being down doesn't lose them), and each tick is capped at 200 records so a backlog can't monopolise the worker.

## Upgrade path

Two patches, because neither case is covered by shipping the fields alone:

- `add_next_follow_up_to_layouts` — the layout seeder in `install.py` is skip-if-exists, so existing sites would get the field and its reminders with nowhere to set the date
- `set_follow_up_reminder_defaults` — field defaults only apply when a Single is first created, so existing sites would otherwise get `0`/`NULL`: reminders off and lead time zero

## Testing

14 integration tests in `crm/tests/test_follow_up.py` covering lead-time maths, due selection, no-duplicate-send, reschedule/clear/closed/converted, the disabled setting, and recipient resolution including the owner fallback and disabled users.

Verified manually in a local bench: setting the date from the side panel, the scheduler producing a real in-app notification, the new list column, and the settings controls persisting.

One bug worth calling out, since the fix looks like it could be simplified away. Frappe's query builder compiles `<=` on a Datetime into `IFNULL(next_follow_up, '0001-01-01') <= cutoff`, so **every record with no follow-up date matched** — the first scheduler tick would have reminded the owner of every lead and deal in the system. `get_due_records` therefore carries an explicit `["next_follow_up", "is", "set"]` filter, with a comment and a regression test (`test_record_without_follow_up_is_never_due`) guarding it.

### DIFF
--- a/crm/api/follow_up.py
+++ b/crm/api/follow_up.py
@@ -59,13 +59,13 @@ def trigger_follow_up_reminders():
 	# instead of leaving records marked as reminded but never notified.
 	for doctype in FOLLOW_UP_DOCTYPES:
 		for record in get_due_records(doctype, cutoff):
-			try:
-				send_reminder(doctype, record, send_email=send_email)
-			except Exception:
-				frappe.log_error(title=f"Follow up reminder failed for {doctype} {record.name}")
-			finally:
-				# Marked either way: a reminder that failed to send is logged, but
-				# retrying it on every tick would just repeat the same failure.
+			# send_reminder isolates each recipient, so what reaches here is a
+			# failure to work out *who* to notify at all -- nothing was sent, and
+			# the flag stays clear so the next tick tries the record again.
+			if run_in_savepoint(
+				lambda record=record: send_reminder(doctype, record, send_email=send_email),
+				title=f"Follow up reminder failed for {doctype} {record.name}",
+			):
 				frappe.db.set_value(doctype, record.name, "follow_up_reminder_sent", 1, update_modified=False)
 
 
@@ -117,7 +117,13 @@ def get_due_records(doctype, cutoff):
 
 
 def send_reminder(doctype, record, send_email=False):
-	"""Notify everyone responsible for ``record`` that its follow up is due."""
+	"""Notify everyone responsible for ``record`` that its follow up is due.
+
+	The record is marked reminded once, so every recipient gets their own
+	savepoint: one user whose notification blows up must not swallow the
+	reminder for the others, and must not poison the surrounding transaction.
+	Whoever failed is left in the Error Log rather than retried forever.
+	"""
 	recipients = get_recipients(doctype, record)
 	if not recipients:
 		return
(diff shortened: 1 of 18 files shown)


chinmaybhatk:
Screenshots for ref:

Screenshot 2026-08-14 at 1 36 54 PM

Screenshot 2026-08-14 at 1 37 30 PM

ps173:
Hey, we are planning on re-thinking the whole follow-up on deals and leads with a different UX. I like the idea that you are putting forward but I don't think this is the way we want to go. Because adding a new field to tell "when next follow up should be" isn't something that should be done on the deal/lead level. Rather it should be something in the tasks and then we can have a better task management that tells you about all the follow-ups that you create. This is definitely not the way to go though. 

CC - @michellealva

chinmaybhatk:
Thanks for sharing your view point @ps173  .
When agent/executive are on a call/working on a lead/deal its easy for the executive to mark when is the next follow up date rather than creating tasks with multiple clicks. And it has become habit for most of the sales people to mark it against the lead either on excelsheet or other CRM in the market which they are used to. 

My advise is, lets ride on the existing habit of users instead of introducing a new flow which will need a training on how to use the feature.

ps173:
> My advise is, lets ride on the existing habit of users instead of introducing a new flow which will need a training on how to use the feature.

We will keep this in mind when planning the new flow. I will cherry-pick and patch some of your commits I land on doing the same thing. Nonetheless as of now I don't think this is the best solution for letting users add a reminder.

PR #2673, opened 2026-08-14
fix(a11y): add alt text to images missing it

swe-sanad:
## Problem

Several images across the frontend had no `alt` attribute.

## Fix

Per-image, not a blanket fill:
- Icons shown next to their own visible text label (app menu, theme preview) get `alt=""` — the label already names them; a real alt would just double-announce the same text to a screen reader.
- Standalone images with no adjacent label get a real description: `BrandLogo.vue` (org brand mark), `AttachmentItem.vue` / `WhatsAppArea.vue` (attachment previews — alt from the filename/label already in scope), `EmailProviderIcon.vue` (falls back to a generic description when its optional label prop is empty).

`Apps.vue` no longer exists on `develop` — its app-menu icon moved into `UserDropdown.vue`'s `appMenuItems()`, fixed there instead.

### DIFF
--- a/frontend/src/components/Activities/WhatsAppArea.vue
+++ b/frontend/src/components/Activities/WhatsAppArea.vue
@@ -97,6 +97,7 @@
           <div v-else-if="whatsapp.content_type == 'image'">
             <img
               :src="whatsapp.attach"
+              alt="WhatsApp attachment"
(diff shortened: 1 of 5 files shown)


ps173:
@swe-sanad please fix the failing pre-commit checks

Issue #401, opened 2024-10-06
Issues while navigating list view in CRM


nehasan16:
When working on a list of leads, I faced the following issues which made my experience really poor. 

1. I open a deal, address it, and press "back"
<img width="1120" alt="Screenshot 2024-10-06 at 3 03 34 PM" src="https://github.com/user-attachments/assets/6b7be312-86fc-47ab-adf9-1f36b29e776b">

2. After pressing back, I lost track of which lead I had really opened and which was the next in line. :( 

Ideally, once I am back to the list view, the view before opening the deal should persist, and the deal I just opened should is clearly visible. In ERPNext this is achieved by making the name of then opportunity "unbolded". In email, this is achieved by giving a different color to unopened emails and a different color to opened emails. 

Basically, currently, it is very hard to go through a list of deals easily. The UX can be much better.

NehaSan16:
@shariquerik Can we please check this?

michellealva:
@nehasan16 the view is persistent, I just checked. However, the bold text one can be done. Can you confirm the view thing if it works for you?

PR #2693, opened 2026-08-18
feat: dim visited deals/leads in list view using frappe's _seen field

Shindhu-Ramaswamy:
issue : #401 
Dims deal/lead rows in the list view once the current user has opened them, using Frappe's built-in `_seen` tracking (the same mechanism desk's own list view uses to bold/unbold rows) instead of a client-side store.

### DIFF
--- a/frontend/src/composables/useVisitedRecords.ts
+++ b/frontend/src/composables/useVisitedRecords.ts
@@ -1,16 +1,16 @@
 import { call } from 'frappe-ui'
 import { sessionStore } from '@/stores/session'
 
-export function useVisitedRecords(doctype) {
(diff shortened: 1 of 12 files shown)


Shindhu-Ramaswamy:
[Screencast from 2026-08-18 11-51-00.webm](https://github.com/user-attachments/assets/2c5f0dfc-04cd-4732-b965-31685c935150)

frontend/src/composables/useVisitedRecords.ts
ps173, line 1: no js file please, typescript first

ps173 (changes requested):

ps173 (approved):

PR #2699, opened 2026-08-19
fix: pin primary contact to top of Deal contacts list

### LINKED ISSUE #2698: Primary contact is not pinned to the top of the Contacts list on a Deal
Current behaviour
On a Deal, the Contacts section lists contacts in the order they were added. A contact marked as primary keeps its original position in that list, so on a deal with several contacts the primary one can be buried at the bottom.

https://github.com/user-attachments/assets/f730e1c8-0757-4c12-ae04-679e89942ce2

Shindhu-Ramaswamy:
Order CRM Contacts by is_primary before idx so the primary contact is always first, instead of keeping its original insertion position.

Fixes #2698

### DIFF
--- a/crm/fcrm/doctype/crm_deal/api.py
+++ b/crm/fcrm/doctype/crm_deal/api.py
@@ -7,6 +7,7 @@ def get_deal_contacts(name: str):
 		"CRM Contacts",
 		filters={"parenttype": "CRM Deal", "parent": name},
 		fields=["contact", "is_primary"],
+		order_by="is_primary desc, idx asc",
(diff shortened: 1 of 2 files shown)


Shindhu-Ramaswamy:
[Screencast from 2026-08-19 17-20-11.webm](https://github.com/user-attachments/assets/9736f09d-5af9-495b-9ea2-9ff5a2acb17b)

ps173 (approved):
Looks a bit abrupt, but fixes things nonetheless. Can improve the UX for this later on

PR #2700, opened 2026-08-19
fix: preserve scroll position in grouped list view

Shindhu-Ramaswamy:
Fixes scroll position not being preserved when using the "Group By" list view. Currently, if you scroll down a grouped list, open a record, and go back, the list jumps back to the top instead of staying where you were — this already works correctly in the normal flat list view, just not the grouped one. This fix makes it work the same way in both.

Related to #401.

### DIFF
--- a/frontend/src/components/ListViews/ListRows.vue
+++ b/frontend/src/components/ListViews/ListRows.vue
@@ -1,5 +1,9 @@
 <template>
-  <div v-if="showGroupedRows" class="mx-3 mt-2 h-full overflow-y-auto sm:mx-5">
+  <div
+    v-if="showGroupedRows"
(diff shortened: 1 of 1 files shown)


Shindhu-Ramaswamy:
Before : [Screencast from 2026-08-19 21-02-42.webm](https://github.com/user-attachments/assets/23703550-1e87-4b0a-9fab-4c983cb64186)

Shindhu-Ramaswamy:
After :  [Screencast from 2026-08-19 21-06-57.webm](https://github.com/user-attachments/assets/d7f2827d-3bd8-4d1a-8df2-e12eb66562aa)

ps173 (approved):
LGTM

PR #2711, opened 2026-08-24
fix:  kanban three-dot click -In Task Page 

### LINKED ISSUE #2707: Kanban card three-dot menu opens Delete and the edit modal simultaneously, and delete has no confirmation
In the Kanban view (Tasks page, and also reproducible on Leads/Deals), clicking the three-dot (⋯) icon on a card does two things at once instead of one:

Opens the dropdown showing "Delete" (and other quick actions on Leads/Deals).
Also opens the record's edit modal, because the click event isn't actually being stopped — it bubbles up to the card and triggers the card's own click handler.
On top of that, clicking "Delete" on the Tasks page deletes the task immediately with no confirmation step

Shindhu-Ramaswamy:
Fixes the Kanban Task card's three-dot (⋯) menu:

Clicking it used to open the task's edit modal and the dropdown at the same time — now it just opens the dropdown.
Delete no longer removes the task instantly; it asks for confirmation first, same as the List view.
Added an explicit Edit option to the menu, so Edit and Delete are both there instead of Edit only working by clicking the card itself.
closes : #2707

### DIFF
--- a/frontend/src/pages/Tasks.vue
+++ b/frontend/src/pages/Tasks.vue
@@ -146,9 +146,12 @@
           class="flex items-center gap-2"
           :options="actions(itemName)"
           variant="ghost"
-          @click.stop.prevent
         >
-          <Button icon="lucide-more-horizontal" variant="ghost" />
+          <Button
+            icon="lucide-more-horizontal"
+            variant="ghost"
+            @click.stop.prevent
+          />
         </Dropdown>
       </div>
     </template>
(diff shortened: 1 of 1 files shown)


Shindhu-Ramaswamy:
[Screencast from 2026-08-24 11-25-02.webm](https://github.com/user-attachments/assets/9907348f-72a2-46c8-a586-07dd45b52d1e)

ps173 (approved):

PR #2717, opened 2026-08-26
fix: render Data field URLs as clickable links in side panel

### LINKED ISSUE #1861: Make URL fields clickable and open in a new tab
**Problem**
Currently, users must copy-paste URLs from Data fields to open them, which is inconvenient.

**Expected Behavior**

- If a Data field contains a valid URL:
- It should be clickable
- It should open in a new tab

Image

Shindhu-Ramaswamy:
Read-only Data fields containing a URL now render as a clickable <a> link (opens in a new tab) instead of plain text. Added a getSafeWebsiteUrl check to validate the value before turning it into a link, and suppressed the separate external-link icon for these fields since it's now redundant.

closes : #1861


ps173:
@Shindhu-Ramaswamy screenshots please

PR #2726, opened 2026-08-27
fix: prevent 'Document not found' error flash on lead delete

### LINKED ISSUE #2718: Deleting a Lead shows "Document not found" error instead of a success message
Describe the bug
When deleting a Lead from its detail page, the deletion actually succeeds, but instead of a success message, the app briefly shows a "Document not found" error page before navigating to the Leads list. It feels like the delete failed even though it worked.

To Reproduce
Open any Lead's detail page.
Click the delete (trash) icon → confirm delete in the popup.
Observe: the lead gets deleted, but a "Document not found" error flashes instead of a success message.

Expected behavior
After a successful delete, the app should show a success message (e.g. "Lead deleted successfully") and navigate cleanly to the Leads list — no error should appear.

https://github.com/user-attachments/assets/8b24d1aa-45ea-44e0-8474-fbd4993b9125

Shindhu-Ramaswamy:
closes: #2718
## Problem

Deleting a Lead /deal(or any document via `DeleteLinkedDocModal`) succeeds, but
a "Document not found" error flashes on screen before the app navigates
away to the list view. The delete actually works — the error is just
misleading noise.

## Fix
Mark the doc as "intentionally deleted" before sending the delete request,
so that expected error gets silently ignored instead of shown to the user.
The marker is cleared once the delete finishes (or immediately if the
delete itself fails), so it doesn't accidentally hide unrelated errors later.

### DIFF
--- a/frontend/src/components/DeleteLinkedDocModal.vue
+++ b/frontend/src/components/DeleteLinkedDocModal.vue
@@ -131,9 +131,10 @@
 </template>
 
 <script setup>
-import { createResource, call } from 'frappe-ui'
+import { createResource, call, toast } from 'frappe-ui'
 import { useRouter } from 'vue-router'
 import { computed, ref } from 'vue'
+import { markDocumentAsDeleted } from '@/data/document'
 
 const show = defineModel({ type: Boolean })
 const router = useRouter()
@@ -253,6 +254,9 @@ const deleteDoc = async () => {
     doctype: props.doctype,
     name: props.docname,
   })
+  markDocumentAsDeleted(props.doctype, props.docname)
+  toast.success(__('{0} deleted successfully', [props.doctype]))
+  show.value = false
   router.push({ name: props.name })
   props?.reload?.()
 }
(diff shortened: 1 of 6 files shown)


Shindhu-Ramaswamy:
[Screencast from 2026-08-28 12-50-59.webm](https://github.com/user-attachments/assets/0ee4b344-d993-49b0-884e-bf7416d7cdf8)

Shindhu-Ramaswamy:
[Screencast from 2026-08-31 13-20-25.webm](https://github.com/user-attachments/assets/8eeab8c7-8f25-4346-ad0a-d3a294e6e931)

Shindhu-Ramaswamy:
updated the delete success toast to include the document's name and title instead of just the doctype label (e.g. CRM Lead deleted successfully) — it now shows something like CRM-LEAD-2026-00001 (Jane Doe) deleted successfully / CRM-DEAL-2026-00001 (Acme Corp) deleted successfully, on both desktop and mobile views.

PR #2739, opened 2026-08-29
fix: use system time format for WhatsApp message timestamps

### LINKED ISSUE #2695: WhatsApp message timestamp hardcoded to 12-hour AM/PM format, ignoring locale/System Settings time format
## Description

In `frontend/src/components/Activities/WhatsAppArea.vue`, the message timestamp is rendered with a hardcoded format string:

```js
{{ formatDate(whatsapp.creation, 'hh:mm a') }}
```

`hh:mm a` forces a 12-hour clock with an AM/PM marker, regardless of:
- the site's System Settings `time_format` (e.g. set to `HH:mm:ss`, i.e. 24-hour)
- the logged-in user's `language` (e.g. `it`, where a 24-hour clock without AM/PM is the norm)

The tooltip on the same line also hardcodes an English date format (`ddd, MMM D, YYYY`), independent of the user's locale.

## Impact

Users on sites configured for a 24-hour clock (or non-English locales) still see WhatsApp message times as e.g. `02:30 pm` instead of `14:30`, which is inconsistent with the rest of the Frappe/CRM UI that does respect System Settings / user locale.

## Suggested fix

Use a locale/System-Settings-aware format instead of the hardcoded `'hh:mm a'`, e.g. deriving the pattern from `frappe.boot.sysdefaults.time_format` (24h vs 12h) the same way other parts of the desk/CRM UI already do, rather than hardcoding a 12-hour US-style format.

## Environment

- frappe/crm: `develop` branch, commit `177a781` (merge of #2615)

Kie70:
Closes #2695

WhatsApp message bubbles hardcode `hh:mm a`, so sites configured with a 24-hour time format still see 12-hour timestamps. List views already render Datetime cells through `getFormat()` / `window.sysdefaults.time_format` (e.g. `Leads.vue`), making the WhatsApp area the odd one out. This switches the bubble to the same mechanism, so the timestamp follows System Settings.

Note: `TaskArea.vue` / `TasksListView.vue` have the same hardcoding; left out to keep this scoped to the reported issue — happy to follow up.

### DIFF
--- a/frontend/src/components/Activities/WhatsAppArea.vue
+++ b/frontend/src/components/Activities/WhatsAppArea.vue
@@ -140,7 +140,7 @@
           <div class="-mb-1 flex shrink-0 items-end gap-1 text-ink-gray-5">
             <Tooltip :text="formatDate(whatsapp.creation, 'ddd, MMM D, YYYY')">
               <div class="text-2xs">
-                {{ formatDate(whatsapp.creation, 'hh:mm a') }}
+                {{ formatDate(whatsapp.creation, '', false, true) }}
               </div>
             </Tooltip>
             <div v-if="whatsapp.type == 'Outgoing'">


ps173:
Won't be going ahead with this as we have a new integration in place #2244

ps173:
> Note: TaskArea.vue / TasksListView.vue have the same hardcoding; left out to keep this scoped to the reported issue — happy to follow up.

Can follow up with this but I am closing this pr

PR #2740, opened 2026-08-29
fix: wrap hardcoded user-facing strings in __() for translation

Kie70:
A pass over the frontend for user-visible strings that never reach translators because they aren't wrapped in `__()` (the extractor only picks up literal first arguments):

- error banners on the Deal/Lead pages (wrapped the inner literals, same pattern as `MobileDeal.vue`)
- `DoctypeModal` create/update fallback errors (the modal even calls `__(error)` at display time, but these msgids were never extracted)
- Table MultiSelect validation messages (also corrects `atleast` → `at least`)
- file-upload errors, empty-state titles/descriptions for all Activities tabs, event-notification fallback title, address-state placeholder, and the `Set` / `Not Set` filter labels

~34 strings across 9 files; pure `__()` wraps plus prettier reflow, no behavior change. Deliberately avoids the files touched by #2432 (complementary — those fix display-time translation, these make the literals extractable). Did not regenerate `main.pot` since that lands via the scheduled chore PRs.

### DIFF
--- a/frontend/src/components/Activities/Activities.vue
+++ b/frontend/src/components/Activities/Activities.vue
@@ -733,49 +733,57 @@ const top = computed(() => {
 })
 
 const emptyText = computed(() => {
-  let text = 'No Activities Found'
+  let text = __('No Activities Found')
   if (title.value == 'Emails') {
-    text = 'No Emails Found'
+    text = __('No Emails Found')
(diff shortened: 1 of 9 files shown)


ps173:
Thanks a lot for this!!

PR #2748, opened 2026-08-31
fix: prevent document not found error on delete

### LINKED ISSUE #2718: Deleting a Lead shows "Document not found" error instead of a success message

Ramymurugan:
closes: #2718

## Problem

Deleting a Lead or Deal succeeds, but a "Document not found" error briefly appears before navigating to the list view.

## Fix

Mark the document as intentionally deleted before the delete request. If the document resource receives the expected `DoesNotExistError` during the deletion flow, the error is ignored instead of being displayed to the user.

The deletion marker is cleared when the delete fails and expires after a short period to prevent it from affecting unrelated errors.

## Testing

* Created a test Lead
* Deleted the Lead successfully
* Verified the "Document deleted successfully" message
* Verified navigation to the Leads list
* Verified the deleted Lead was removed
* Verified the "Document not found" error was not displayed

### DIFF
--- a/frontend/src/components/DeleteLinkedDocModal.vue
+++ b/frontend/src/components/DeleteLinkedDocModal.vue
@@ -131,9 +131,10 @@
 </template>
 
 <script setup>
-import { createResource, call } from 'frappe-ui'
+import { createResource, call, toast } from 'frappe-ui'
 import { useRouter } from 'vue-router'
 import { computed, ref } from 'vue'
+import { deletedDocuments } from '@/data/document'
 
 const show = defineModel({ type: Boolean })
 const router = useRouter()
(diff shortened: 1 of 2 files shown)


ps173:
This was resolved in #2726

PR #2749, opened 2026-08-31
Make website column clickable in Leads and Deals list views

### LINKED ISSUE #2750: Website column in Leads/Deals list view is not clickable
Description:

In the Leads and Deals list views, the Website column displays the website value as plain text. Users have to manually copy the value and open it in a new tab instead of clicking through directly.

Steps to reproduce:

Go to CRM → Leads (or Deals) list view.
Add the Website column if not already visible.
Note that the website value is rendered as plain, non-interactive text.
Expected behavior:
Clicking the website value in the list view should open the URL in a new tab, the same way it works elsewhere in the app.
Actual behavior:
The value is static text with no link/click behavior.

Shindhu-Ramaswamy:
closes : #2750 

Makes the website field in Leads and Deals list views clickable — it now opens the URL in a new tab instead of showing as plain text.

### DIFF
--- a/frontend/src/components/ListViews/DealsListView.vue
+++ b/frontend/src/components/ListViews/DealsListView.vue
@@ -177,6 +177,14 @@
                 })
             "
           />
+          <button
+            v-else-if="column.key === 'website' && item?.url"
+            type="button"
+            class="w-full truncate bg-transparent p-0 text-left text-base cursor-pointer hover:underline focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-outline-gray-4 focus-visible:ring-offset-1 rounded"
+            @click.stop.prevent="openWebsite(item.url)"
+          >
+            {{ getLabel(label, column) }}
+          </button>
           <div
             v-else-if="label"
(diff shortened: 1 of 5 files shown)


Shindhu-Ramaswamy:
[Screencast from 2026-08-31 22-03-27.webm](https://github.com/user-attachments/assets/1fba7f90-e67b-4166-a46f-c198b290a624)

frontend/src/components/ListViews/DealsListView.vue
```diff
@@ -197,6 +202,14 @@
...
           </div>
         </template>
+        <template #suffix>
+          <span
+            v-if="column.key === 'website' && item?.url"
+            class="lucide-external-link h-3.5 w-3.5 shrink-0 cursor-pointer text-ink-gray-5"
+            aria-hidden="true"
+            @click.stop.prevent="openWebsite(item.url)"
+          />
+        </template>
```
ps173: repeating logic, probably extract this somewhere

ps173 (changes requested):

PR #2751, opened 2026-09-01
fix: propagate no_of_employees to organization on lead/deal conversion

### LINKED ISSUE #2744: Organization default employee count should respect the source
Currently the default employee count is 1-10. It should just pick the count from the lead / deal that is linked to it

Shindhu-Ramaswamy:
The employee count entered on a Lead or Deal was getting lost when it was converted into an Organization — it either wasn't copied at all, or got silently reset to the default 1-10. This fix passes no_of_employees through in both conversion paths so the value carries over correctly.

closes : #2744


ps173 (approved):
LGTM!

PR #2754, opened 2026-09-01
fix: persist details bar width across page refresh

### LINKED ISSUE #2040: Details bar width changes after refresh
https://github.com/user-attachments/assets/6f3c16db-66ed-47d9-9239-048caa28c811

#### Version
- crm: 1.66.1

Shindhu-Ramaswamy:
closes :  #2040

### DIFF
--- a/frontend/src/components/Resizer.vue
+++ b/frontend/src/components/Resizer.vue
@@ -24,7 +24,19 @@ const props = defineProps({
 })
 
 const sidebarResizing = ref(false)
-const sidebarWidth = ref(props.defaultWidth)
+const sidebarWidth = ref(getInitialWidth())
+
+function getInitialWidth() {
+  let storedWidth = Number(localStorage.getItem('sidebarWidth'))
+  if (
+    storedWidth &&
+    storedWidth >= props.minWidth &&
+    storedWidth <= props.maxWidth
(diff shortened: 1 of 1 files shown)


Shindhu-Ramaswamy:
[Screencast from 2026-09-01 14-40-47.webm](https://github.com/user-attachments/assets/68a2f671-c6a8-4945-9d1a-39f3a13ce0b9)

frontend/src/components/Resizer.vue
```diff
@@ -24,7 +24,19 @@ const props = defineProps({
...
+  let storedWidth = Number(localStorage.getItem('sidebarWidth'))
+  if (
+    storedWidth &&
+    storedWidth >= props.minWidth &&
+    storedWidth <= props.maxWidth
+  ) {
+    return storedWidth
+  }
+  return props.defaultWidth
+}
```
ps173, line 40: We can probably save this in layout somewhere. Cause everything about layout is saved in the custom view doctype. I think this should be as well. 

Probably should find a way to extend CRM View Settings

ps173: @shariquerik clarified that this is not required. CRM view settings are for site wide things and not the user level settings

PR #2765, opened 2026-09-02
fix: patch for marking all existing records as seen

ps173:


ps173:
Duplicated https://github.com/frappe/crm/pull/2767/

PR #2767, opened 2026-09-02
fix: dim assignee, timestamp and favourite columns for visited rows

Shindhu-Ramaswamy:
## Summary
Follow-up to #2693. @ps173 pointed out that the visited-row dimming didn't carry to every column, and that a new `track_seen` field needs a data patch for existing customer records.

- The `isVisited` dim styling was only wired into the generic text fallback branch in `DealsListView.vue`/`LeadsListView.vue`, so the assignee avatar's name, the timestamp columns (modified/creation/first_response_time/first_responded_on/response_by) and the favourite heart icon stayed bold even after a row was opened. Applied the same `isVisited`-based classes to all three.
- Added a `labelClass` prop to `MultipleAvatar` so the single-assignee name label can pick up the dim/bold styling from the parent.
- Added patch `set_seen_for_existing_deals_and_leads` to backfill `_seen` (as `[owner]`) for CRM Deals/Leads that predate `track_seen` being enabled, so existing records don't all show up as unvisited.

## Test plan
- [x] `yarn build` (frontend) succeeds
- [x] `eslint` on changed Vue files passes clean
- [x] pre-commit hooks (ruff, prettier, eslint, check-python-ast) pass on commit
- [x] Manually verify in the list view: open a deal/lead, confirm assignee name, timestamp column and favourite icon all dim on return to the list
- [x] Run `bench migrate` on a site with existing deals/leads and confirm `_seen` gets backfilled

### DIFF
--- a/crm/patches/v1_0/set_seen_for_existing_deals_and_leads.py
+++ b/crm/patches/v1_0/set_seen_for_existing_deals_and_leads.py
@@ -1,0 +1,46 @@
+import json
+
+import frappe
+from frappe.query_builder import DocType
+
+from crm.api.session import CRM_ALLOWED_ROLES
+
+DOCTYPES = ("CRM Deal", "CRM Lead")
+
+
+def execute():
+	"""Backfill `_seen` for CRM Deals/Leads that predate track_seen being enabled.
+
+	Without this, every record created before track_seen was turned on would
+	show up as unvisited (bold) to everyone, since there's no real view
+	history to reconstruct. Default them to already seen by every enabled
+	CRM user instead, so only records created or updated from now on show
+	as unvisited.
+	"""
+	users = get_crm_users()
+	if not users:
+		return
+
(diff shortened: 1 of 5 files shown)


crm/patches/v1_0/set_seen_for_existing_deals_and_leads.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+def execute():
+	"""Backfill `_seen` for CRM Deals/Leads that predate track_seen being enabled.
+
+	Without this, every record created before track_seen was turned on would
+	show up as unvisited (bold) to everyone, since there's no real view
+	history to reconstruct. Default them to already seen by every enabled
+	CRM user instead, so only records created or updated from now on show
+	as unvisited.
```
ps173: no need to explain what this does. Code explains

PR #2770, opened 2026-09-02
Fix:deal contact organization filter

### LINKED ISSUE #1203: CRM Deal → Contact field shows all Contacts instead of those linked to selected Organization
**Actual behavior:**
Contact field ignores the Organization context and shows all contacts.

**Expected behavior:**
If an Organization is selected in a Deal, the Contact field should only list Contacts linked to that Organization

Image

.

Shindhu-Ramaswamy:
Closes #1203

### Problem

In a Deal, the Contact dropdown listed every contact in the system, ignoring the
Deal's Organization.

The obvious fix is to filter the dropdown by organization. But a contact is tied
to an organization through Contact.company_name, and until now that field could
only be set from the contact's own page — there was no way to do it from the
organization. So filtering alone would leave most users with an empty dropdown
and no way out.

This PR fixes the missing link first, then makes the dropdown helpful without
hiding anything.

### Changes

**1. Add contacts from the Organization page**

The Contacts tab on an organization was read-only. It now has an **Add Contact**
button that either:
- links an existing contact to the organization, or
- opens the contact modal with the organization already filled in ("Create New")

Added on both desktop and mobile.

**2. Group the Deal's contact dropdown instead of filtering it**

The dropdown now splits into two labelled groups — the organization's contacts
first, everyone else below — so the org's contacts are easy to find but no
contact is ever hidden. A deal's contact isn't always someone already linked to
the organization, so removing them from the list would be wrong.

This is done with a new optional `grouping` prop on the `Link` control, which
reuses the grouped-options support already present in `Autocomplete`.

### DIFF
--- a/frontend/src/pages/MobileOrganization.vue
+++ b/frontend/src/pages/MobileOrganization.vue
@@ -129,6 +129,32 @@
           :columns="columns"
           :options="{ selectable: false, showTooltip: false }"
         />
+        <div v-if="tab.label === 'Contacts'" class="flex justify-end px-3 pt-3">
+          <Link
+            value=""
+            doctype="Contact"
+            :filters="{ company_name: ['!=', props.organizationId] }"
+            :onCreate="
+              (value, close) => {
+                _contact = {
+                  first_name: value,
+                  company_name: props.organizationId,
+                }
+                showContactModal = true
+                close()
+              }
+            "
+            @change="(contact) => addContact(contact)"
+          >
+            <template #target="{ togglePopover }">
+              <Button
+                :label="__('Add Contact')"
+                iconLeft="plus"
+                @click="togglePopover()"
+              />
+            </template>
+          </Link>
+        </div>
         <ContactsListView
           v-if="tab.label === 'Contacts' && rows.length"
           class="mt-4"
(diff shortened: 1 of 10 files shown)


Shindhu-Ramaswamy:
Screenshot from 2026-09-02 16-30-40

Shindhu-Ramaswamy:
Screenshot from 2026-09-04 14-58-36
Screenshot from 2026-09-04 14-59-28

frontend/src/pages/MobileOrganization.vue
```diff
@@ -129,6 +129,32 @@
...
+            @change="(contact) => addContact(contact)"
+          >
+            <template #target="{ togglePopover }">
+              <Button
+                :label="__('Add Contact')"
+                iconLeft="plus"
+                @click="togglePopover()"
+              />
+            </template>
+          </Link>
```
ps173, line 156: Why is button wrapped inside of link ?

Shindhu-Ramaswamy: #target is Link's own slot for the element that opens the popover, so putting the Button there just makes the dropdown open from a button instead of a text input.
I followed the pattern deal.vue already uses for adding a contact .

Shindhu-Ramaswamy:
[Screencast from 2026-09-09 15-44-05.webm](https://github.com/user-attachments/assets/9806e6c6-4e99-4c34-9ca2-0a6c3cdf53b4)

PR #2774, opened 2026-09-03
feat: lead segments with bulk add and remove of leads

Neha611:
## Problem

There is no way to group CRM Leads into a persistent, named set. Leads can be filtered ad hoc or converted to deals, but a curated list cannot be saved, revisited, or handed to a teammate.

## What this adds

A **CRM Lead Segment** doctype — segment name, description, an assignee, and a child table of leads — with basic CRUD and bulk add/remove of leads.

### Backend

- `CRM Lead Segment` (`segment_name` unique + autoname, `assigned_to` → User, `description`, `leads` table) and the `CRM Lead Segment Leads` child table.
- `assigned_to` is validated against `CRM_ALLOWED_ROLES` from `crm/api/session.py`, so a segment can only be handed to a CRM user.
- Three whitelisted, permission-checked endpoints: `add_leads`, `remove_leads`, `get_segment_leads`.
- `get_segment_leads` reuses `CRMLead.default_list_data()`, so the leads inside a segment render with the same columns as the main Leads list, and goes through `frappe.get_list` so `org_hierarchy` permission query conditions still apply.
- A patch creates the quick entry layout on existing sites (`add_default_fields_layout` only runs on `after_install`).

### Uniqueness

A lead belongs to a segment **at most once**, but may belong to many segments. This is enforced by a composite unique index on `(parent, lead)`, declared in `on_doctype_update()` — the same mechanism `Email Group Member` uses for `(email_group, email)`. A single-column `unique` on `lead` would have made a lead globally unique and let it join only one segment ever.

`validate()` throws first with a readable message naming the lead, because Frappe's unique-violation handler maps a composite index back to its first column and would otherwise report "Parent must be unique". `add_leads` also filters duplicates up front so one already-present lead doesn't abort an entire batch.

### Frontend

- Sidebar entry, **Lead Segments** list page, and a segment detail page showing its leads with a bulk **Remove from Segment** action.
- **Add to Segment** added to the Leads list bulk actions.
- Create/edit reuses the existing `useDoctypeModal` + `CreateDocumentModal`, so there is no bespoke form.

## Tests

13 backend tests, all passing. They cover the invariant from both sides: a raw `db_insert` of a duplicate row raises `UniqueValidationError` (proving the index is real, independent of `validate`), and the same lead can still be added to two different segments.

## Note on scale

`add_leads`/`remove_leads` go through `doc.save()`, which is O(total members) because `update_child_table` re-issues `db_update()` per row with no dirty check. This is comfortable in the hundreds and degrades in the thousands. Happy to move the writes to direct child-row inserts/deletes with dedupe as an indexed query if you'd prefer that in this PR rather than a follow-up.


rmehta:
Add a video? Doc?

Neha611:
Video Demo -: https://drive.google.com/file/d/1FPbcNL9rSfocstj3E7jnWR2VfXMNJBgK/view?usp=sharing
Doc -: https://claude.ai/code/artifact/c3d29d83-be42-443f-845d-4570e842af01

Neha611:
> Add a video? Doc?

I have added screen recording and document explaining complete feature 
I was facing issue with screen recorder that is why I couldn't voiceover walk through 
Video Demo -: https://drive.google.com/file/d/1FPbcNL9rSfocstj3E7jnWR2VfXMNJBgK/view?usp=sharing
Doc -: https://claude.ai/code/artifact/c3d29d83-be42-443f-845d-4570e842af01

But I can give voice walkthrough too if you want

ps173:
I am having a hard time to understand the use case here. We already have views in existence that provide exact same functionality based on what you are trying to achieve. You can filter the leads that are assigned to you and create a public or private view out of it. Bulk assignment is also a existing feature and on top of it we also have assignment rules in lead and deal both which allow automation of it. 

Another problem with this implementation is that it dies on arrival due user hierarchy. Sales Users can only see the leads that are assigned to them. 

To proceed further with this I would need some sort of clarification on intent as the problems in above description are already solved in current implementation

Neha611:
@ps173 You’re correct that Views can be used to achieve something similar. However, in many real-world scenarios, Views return all leads matching a particular filter, while a salesperson or CRM user may not want to perform a bulk operation—such as an email campaign—on every lead in that result. They may want to select only a specific subset of leads, which is where Lead Segments become useful.

With Views, users would have to repeatedly select the required leads and perform the operation each time. With Lead Segments, they can create the desired group once and directly perform operations on that segment whenever needed.

Another important advantage is that a Lead Segment can include leads from different organizations, territories, etc., as long as they share a common behavior, characteristic, or business requirement. This makes segments useful for targeting and recurring actions that aren't necessarily tied to organizational or geographical filters.

There may be some gaps in my current implementation, but I believe this feature can be quite valuable at scale. I’m happy to work on those gaps and take this feature forward end-to-end.

ps173:
> With Lead Segments, they can create the desired group once and directly perform operations on that segment whenever need
> Another important advantage is that a Lead Segment can include leads from different organizations, territories, etc., as long as they share a common behavior, characteristic, or business requirement. This makes segments useful for targeting and recurring actions that aren't necessarily tied to organizational or geographical filters.

Leads are not contacts. They are meant to be converted to deal. I don't see any application for long term filtering and categorization to them. Also there aren't a lot of bulk actions that we want to allow on leads. We can improve on the bulk-actions allowed in view in that scenario which will subsequently help list view as well. But again I don't agree with this. 

Closing this PR for now

PR #2775, opened 2026-09-03
fix: link Task/Note created from Call Log modal to its Lead/Deal

### LINKED ISSUE #2773: Task/Note created from Call Log modal doesn't appear on the Lead/Deal's Tasks/Notes tab
**Describe the bug**
When creating a Task or Note from the Call Log detail modal (on a Lead or Deal), the new Task/Note does not appear on that Lead/Deal's own Tasks/Notes tab. It only shows up in the global Tasks/Notes list.

**To Reproduce**

Open a Lead or Deal that has a call log.
Open the Call Log detail modal.
Create a Task (or Note) from within that modal.
Go to the Lead/Deal's own Tasks tab (or Notes tab).
The newly created Task/Note is not there.
Check the global Tasks (or Notes) list — it appears there instead.

**Expected behavior**
A Task/Note created from a Lead/Deal's Call Log should be linked back to that Lead/Deal and show up on its Tasks/Notes tab, the same way a Task/Note created directly from the Tasks/Notes tab does.

Shindhu-Ramaswamy:
closes : #2773

Tasks/Notes created from the Call Log detail modal never got a reference_doctype/reference_docname pointing at the call log's Lead/Deal, so they were invisible on that Lead/Deal's own Tasks/Notes tab and only showed up in the global list .

The frontend inserts the Task/Note itself before calling add_note_to_call_log/add_task_to_call_log, so the fix has to set the reference in the "existing document" branch, not the "create new" branch. Resolving the call log's Lead/Deal also has to check the links child table first, since Twilio/Exotel-created call logs only carry that association there, never in reference_doctype/reference_docname directly.

### DIFF
--- a/crm/integrations/api.py
+++ b/crm/integrations/api.py
@@ -72,11 +72,33 @@ def set_default_calling_medium(medium: str):
 	return get_user_default_calling_medium()
 
 
+def _get_call_log_lead_or_deal(call_log):
+	"""Resolve the Lead/Deal a call log currently belongs to.
+
+	Telephony integrations (Twilio, Exotel) link a call log to its Lead/Deal
+	only via the "links" child table (CRMCallLog.link_with_reference_doc) -
+	they never populate reference_doctype/reference_docname directly. So the
+	links table is checked first, falling back to the direct reference
+	fields for call logs that do have them set.
+	"""
+	for link in call_log.get("links") or []:
+		if link.link_doctype in ("CRM Lead", "CRM Deal"):
+			return link.link_doctype, link.link_name
+
+	if call_log.reference_doctype in ("CRM Lead", "CRM Deal"):
+		return call_log.reference_doctype, call_log.reference_docname
+
+	return None, None
+
+
 @frappe.whitelist()
 def add_note_to_call_log(call_sid: str, note: dict):
 	"""Add/Update note to call log based on call sid."""
 	if not frappe.has_permission("CRM Call Log", "write", call_sid):
 		frappe.throw(_("Not permitted"), frappe.PermissionError)
+
+	call_log = frappe.get_cached_doc("CRM Call Log", call_sid)
+	reference_doctype, reference_docname = _get_call_log_lead_or_deal(call_log)
 
 	_note = None
 	if not note.get("name"):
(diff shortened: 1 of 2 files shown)


Shindhu-Ramaswamy:
[Screencast from 2026-09-07 13-46-17.webm](https://github.com/user-attachments/assets/cc5921b3-b3de-4007-88cf-ab5ceca116d1)

crm/integrations/api.py
```diff
@@ -72,25 +72,64 @@ def set_default_calling_medium(medium: str):
...
+	"""Resolve the Lead/Deal a call log currently belongs to.
+
+	Telephony integrations (Twilio, Exotel) link a call log to its Lead/Deal
+	only via the "links" child table (CRMCallLog.link_with_reference_doc) -
+	they never populate reference_doctype/reference_docname directly. The
+	links table is append-only, so if it ever accumulates more than one
+	Lead/Deal entry, the most recently added one wins - mirroring
+	crm_call_log.parse_call_log, which resolves the same "_lead"/"_deal"
+	the UI displays by letting later links overwrite earlier ones.
+	"""
```
ps173: unncessary comments

crm/integrations/api.py
```diff
@@ -72,25 +72,64 @@ def set_default_calling_medium(medium: str):
...
 @frappe.whitelist()
 def add_note_to_call_log(call_sid: str, note: dict):
 	"""Add/Update note to call log based on call sid."""
 	if not frappe.has_permission("CRM Call Log", "write", call_sid):
 		frappe.throw(_("Not permitted"), frappe.PermissionError)
 
+	call_log = frappe.get_cached_doc("CRM Call Log", call_sid)
+	reference_doctype, reference_docname = _get_call_log_lead_or_deal(call_log)
+
 	_note = None
```
ps173: Can we handle this on the frontend instead, in the modal that does the `frappe.client.insert`?

`CallLogDetailModal` already knows the Lead/Deal. We can do something like this in `CallLogDetailModal.vue`:

```js
function referenceDefaults() {
  const data = callLog.value?.data
  if (data?._lead) return { reference_doctype: 'CRM Lead', reference_docname: data._lead }
  if (data?._deal) return { reference_doctype: 'CRM Deal', reference_docname: data._deal }
  return {}
}

// showNote
defaults: referenceDefaults(),
// showTask
defaults: { status: 'Backlog', priority: 'Low', ...referenceDefaults() },
```

With that, the note/task is inserted with the reference already set, so the backend does not need the "existing document" branch changes, the `_get_call_log_lead_or_deal` helper, or the comments explaining that the insert branch is dead for this flow. The tests covering that path can go too.

Issue #436, opened 2024-10-31
Frappe CRM UX feedback


nehasan16:
<ol data-pm-slice="3 1 []"><li><p><strong>Back and forth between leads is not very effective</strong></p></li></ol><p>It isn’t easy to hop from one deal to another. The only way is to go back to the list view and find the next deal in the queue. But even then, we lose track of the sequence because there is no way to know which deal was just opened. ERPNext used to “unbold” the names of the fields after they were just opened which was helpful.</p><p>When a deal is opened, Instead of scrolling down to the last conversation all the time, can we change the sort order to “latest first”, having the screen auto-scroll down to the last conversation is overwhelming. Similarly, when we go back from the deal view to the list view, the list view goes back on top and then auto-scrolls down, instead of a persistent list view.</p><p>Just like email management, these two improvements will help save a lot of cognitive burden. And also relevant for helpdesk <span class="mention" data-type="mention" data-id="ritvik@frappe.io" data-label="Ritvik Sardana">@Ritvik Sardana</span></p>
Frappe CRM | ERPNext / Frappe Framework
-- | --
  |  

[...]

PR #2778, opened 2026-09-04
fix: expand primary contact by default in deal side panel

Shindhu-Ramaswamy:
The deal side panel rendered every contact collapsed, hiding the email and mobile number behind a click on both desktop and mobile. The state was reset on every fetch, so it could not be dismissed by the user either.

get_deal_contacts orders by "is_primary desc, idx asc", so the first entry is the primary contact whenever one is set, and the top contact otherwise. Keying the expanded state off that position always opens exactly one contact, where checking is_primary alone would have left every contact collapsed on deals with no primary set.

Refs #436 (item 4)

### DIFF
--- a/frontend/src/pages/Deal.vue
+++ b/frontend/src/pages/Deal.vue
@@ -729,8 +729,10 @@ const dealContacts = createResource({
   params: { name: props.dealId },
   cache: ['deal_contacts', props.dealId],
   transform: (data) => {
-    data.forEach((contact) => {
-      contact.opened = false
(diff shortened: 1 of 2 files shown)


Shindhu-Ramaswamy:
[Screencast from 2026-09-04 21-19-13.webm](https://github.com/user-attachments/assets/8c115514-cb6c-4d60-a80d-53bb69c2b9da)

Shindhu-Ramaswamy:
Screenshot from 2026-09-04 21-55-22

ps173 (approved):

PR #2780, opened 2026-09-05
fix: map Exotel call statuses to valid Call Log options

### LINKED ISSUE #824: Error while creating/updating call record - null and no-answer
CRM Call Log Doctype - options for Status field should allow null and no-answer.
Current options have No Answer as option, which is not matching with no-answer status from Exotel.

'cmd': 'exotel_integration.handler.handle_request'}
      status = 'no-answer'
      call_log = <CallLog: 2a695c0deb2abdd9a7def37b22981959>
      e = ValidationError(' Status cannot be "no-answer". It should be one of "Ringing", "In Progress", "Completed", "Failed", "Busy", "No Answer", "Queued", "Canceled"')

      out = {'message': ' Status cannot be "null". It should be one of "Initiated", "Ringing", "In Progress", "Completed", "Failed", "Busy", "No Answer", "Queued", "Canceled"', 'title': 'Message', 'indicator': 'red', 'raise_exception': 1, '__frappe_exc_id': '0c6c9e37b011166af82c37e2cf56aa3f0ad10afe3fdc69144a2cc204'}
  File "apps/frappe/frappe/__init__.py", line 525, in _raise_exception
    raise exc
      exc = ValidationError(' Status cannot be "null". It should be one of "Initiated", "Ringing", "In Progress", "Completed", "Failed", "Busy", "No Answer", "Queued", "Canceled"')
      inspect = <module 'inspect' from '/usr/lib/python3.11/inspect.py'>
      msg = ' Status cannot be "null". It should be one of "Initiated", "Ringing", "In Progress", "Completed", "Failed", "Busy", "No Answer", "Queued", "Canceled"'
      out = {'message': ' Status cannot be "null". It should be one of "Initiated", "Ringing", "In Progress", "Completed", "Failed", "Busy", "No Answer", "Queued", "Canceled"', 'title': 'Message', 'indicator': 'red', 'raise_exception': 1, '__frappe_exc_id': '0c6c9e37b011166af82c37e2cf56aa3f0ad10afe3fdc69144a2cc204'}
      raise_exception = <class 'frappe.exceptions.ValidationError'>
frappe.exceptions.ValidationError:  Status cannot be "null". It should be one of "Initiated", "Ringing", "In Progress", "Completed", "Failed", "Busy", "No Answer", "Queued", "Canceled"

Trade ID 
no-answer 51d210b8-b90c-4646-bb25-6e277eb4e0d0
null 568127fa-cd5a-4984-84ff-80157220227d

Shindhu-Ramaswamy:
Fixes #824

## Problem

Exotel reports call status in its own vocabulary (no-answer, in-progress,canceled, and sometimes the literal string null when a leg never connected). get_call_log_status() only mapped a handful of CallType + status combinations
and returned the raw Exotel string on fall-through. CRM Call Log status is a mandatory Select, so the save failed with Status cannot be "no-answer". It should be one of "Ringing", "In Progress",..and the call log was never created — the call did not show up in CRM at all.

## Fix

- Add EXOTEL_CALL_STATUSES, a full map from Exotel's statuses to the Select options, and normalize_call_status() which lowercases/trims before lookup.
- Treat "", "null", "none","undefined" as *no status* rather than an  unknown one — Exotel documents Leg2Status as empty when there's no second leg.
  In that case the existing status on the call log is left untouched instead of being overwritten with an invalid value.
- Keep the existing overrides: CallType=completed → Completed, busy → Ringing (still ringing on another leg).
- create_call_log() defaults to Ringing if no status is supplied.

### DIFF
--- a/crm/integrations/exotel/handler.py
+++ b/crm/integrations/exotel/handler.py
@@ -210,7 +210,7 @@ def create_call_log(
 	call_log.to = to_number
 	call_log.medium = medium
 	call_log.type = call_type
-	call_log.status = status
+	call_log.status = status or "Ringing"
 	call_log.telephony_medium = "Exotel"
 	setattr(call_log, "from", from_number)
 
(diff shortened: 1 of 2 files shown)


crm/integrations/exotel/handler.py
```diff
@@ -210,7 +210,7 @@ def create_call_log(
 	call_log.to = to_number
 	call_log.medium = medium
 	call_log.type = call_type
-	call_log.status = status
+	call_log.status = status or "Ringing"
```
ps173: If there is no status it defaults to ringing ? this doesn't seem correct

crm/integrations/exotel/handler.py
```diff
@@ -248,32 +248,56 @@ def get_call_log(call_payload):
...
+def normalize_call_status(status):
+	"""Map a raw Exotel status onto a CRM Call Log status, or None if there isn't one."""
+	if not status:
+		return None
+
+	status = str(status).strip().lower().replace("_", "-").replace(" ", "-")
+	if status in EXOTEL_EMPTY_STATUSES:
+		return None
+
+	return EXOTEL_CALL_STATUSES.get(status)
```
ps173: An unknown non-empty status means this mapping table is stale, which is the exact bug this PR fixes. Silently returning None here hides that. Log it so it shows up in Error Log, still return None so the create path falls back and the update path keeps the last good value.

```suggestion
	if status not in EXOTEL_CALL_STATUSES:
		frappe.log_error(
			title="Unknown Exotel call status",
			message=f"Exotel sent call status {status!r}, which is not in EXOTEL_CALL_STATUSES",
		)
		return None

	return EXOTEL_CALL_STATUSES[status]
```

crm/tests/test_integrations.py
```diff
@@ -496,6 +501,105 @@ def test_get_contact_lead_or_deal_from_number_ignores_converted_leads(self):
...
+		self.assertEqual(normalize_call_status(" Completed "), "Completed")
+		self.assertEqual(normalize_call_status("cancelled"), "Canceled")
+
+	def test_normalize_call_status_treats_empty_values_as_unset(self):
+		"""A leg that never connected has no status, rather than an unknown one"""
+		for status in (None, "", "null", "NULL", "undefined"):
+			self.assertIsNone(normalize_call_status(status))
+
+	def test_normalize_call_status_returns_none_for_unknown_status(self):
+		self.assertIsNone(normalize_call_status("some-new-exotel-status"))
```
ps173: Covers the logging half of the change above.

```suggestion
	def test_normalize_call_status_logs_unknown_status(self):
		"""An unknown status means the mapping is stale, so leave a trace instead of guessing"""
		frappe.db.delete("Error Log", {"method": "Unknown Exotel call status"})

		self.assertIsNone(normalize_call_status("some-new-exotel-status"))

		error_logs = frappe.get_all(
			"Error Log", filters={"method": "Unknown Exotel call status"}, pluck="error"
		)
		self.assertEqual(len(error_logs), 1)
		self.assertIn("some-new-exotel-status", error_logs[0])
```

PR #2786, opened 2026-09-07
fix: show quick filters in mobile list view

### LINKED ISSUE #1768: Show filters in lead/deals in mobile view
Currently in mobile view, filters fields are not displayed

Image

Image

Shindhu-Ramaswamy:
Quick filter fields were rendered only in the desktop branch of ViewControls, so on mobile the list toolbar showed just the Filter popover and the sort/columns buttons. Filtering by a standard field took four taps instead of one.

The data was already there — get_quick_filters was being fetched on mobile and quickFilterList computed correctly. Only the markup was missing.

This adds the quick filter row to the mobile branch, laid out the same way as desktop: a horizontally scrollable field list on the left, divider, then the toolbar buttons on the right.

Fixes #1768

### DIFF
--- a/frontend/src/components/ViewControls.vue
+++ b/frontend/src/components/ViewControls.vue
@@ -4,13 +4,29 @@
     class="flex flex-col justify-between gap-2 sm:px-5 px-3 py-4"
   >
     <div class="flex flex-col gap-2">
-      <div class="flex items-center justify-between gap-2 overflow-x-auto">
-        <div class="flex gap-2">
-          <Filter
-            v-model="list"
-            :doctype="doctype"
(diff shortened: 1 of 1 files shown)


ps173:
Attach screenshots please

Shindhu-Ramaswamy:
[Screencast from 2026-09-09 17-47-24.webm](https://github.com/user-attachments/assets/77503abf-1c3d-4318-833e-49efeea150f3)

shahzeelahmed:
@Shindhu-Ramaswamy this should be a "fix" [ref](https://www.conventionalcommits.org/en/v1.0.0/)

Issue #2790, opened 2026-09-07
Calendar/Event feature partially deleted on release branch, dangling imports in CalendarEventPanel.vue, EventModal.vue, EventArea.vue, EventNotificationPopup.vue


Kirito70:
**Describe the bug**
Every recent tagged release (confirmed on `v1.83.0`, the latest at time of writing) ships four frontend files whose imports point at files that don't exist in the tree:

- `frontend/src/components/Calendar/CalendarEventPanel.vue` → imports `Attendee.vue`, `EventNotifications.vue`, `ShortcutTooltip.vue`, and 5 icon components (`BellIcon`, `MapIcon`, `ShieldIcon`, `PeopleIcon`, `DescriptionIcon`) none of which exist
- `frontend/src/components/Modals/EventModal.vue` → imports `Calendar/EventNotifications.vue`, `Calendar/Attendee.vue` missing
- `frontend/src/components/Activities/EventArea.vue` → imports `@/composables/event` —> missing
- `frontend/src/components/EventNotificationPopup.vue` → imports `@/data/notifications` —>  missing

These 4 files are currently unreachable from the app entry point (nothing else in `frontend/src` imports `EventArea.vue`, `EventModal.vue`, `EventNotificationPopup.vue`, or `CalendarEventPanel.vue`), so `yarn build` doesn't fail today but any static import-resolution tooling, or any future change that wires these components back in, breaks immediately.

**To Reproduce**
1. `git clone --branch v1.83.0 --depth 1 https://github.com/frappe/crm.git`
2. `grep -n "^import" frontend/src/components/Calendar/CalendarEventPanel.vue`
3. See imports referencing files not present anywhere in the checkout (e.g. `@/components/Calendar/Attendee.vue`, `@/components/Icons/BellIcon.vue`)

**Expected behavior**
Every file present in the shipped release should have its imports resolvable either the full Calendar/Event feature (event scheduling with attendees + reminder notifications) should be present, or these 4 orphaned files shouldn't be in the tree at all.

**Screenshots**
N/A —> this is a static source/build issue, not a runtime UI bug.

**Desktop / Smartphone**
N/A —> reproducible from the source tree directly, not OS/browser/device-specific.

**Additional context**
Root-caused this via git history on `main-hotfix`:

- The Calendar/Event feature was built correctly on `develop` across several 2025 PRs.
[...]

PR #2816, opened 2026-09-11
chore: remove calendar feature from frontend

ps173:
Ref: #2790 

removes everything related to calendar and event feature


ps173:
Checked for persisted Calendar route references, there are none. Notification rows only route to Lead or Deal, saved views derive their route from the doctype, and the last-open tab store falls back to the first tab if it holds `events`.

PR #2821, opened 2026-09-11
fix: remove duplicate error toast on lead and deal field update

### LINKED ISSUE #1703: Multiple Duplicate Error Messages When Server Script Validation Fails on before save in CRM Lead
# Bug Report: Duplicate Error Toast Notifications on Server Script Validation Failure

## Describe the bug

When a Server Script with validation logic (using `validate` or `before_save` event) throws an error during CRM Lead save operation, the same error message appears **3 times** as toast notifications in the UI, even though only a **single field** is being updated.

### 1. Create a Server Script

- Go to **Server Script List** → Click **New**
- Set **Script Type:** `Document Event`
- Set **Reference DocType:** `CRM Lead`
- Set **Document Event:** `Validate` (or `Before Save`)
- Add the following script:

u can add a simple frappe.throw u will get three times
```
        frappe.throw("Editing this lead is not allowed.")

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/ebbe3969-8295-4e06-a0f4-81f335460a49" />

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/08d03f1f-c553-48c1-bc53-d80eb31ec693" />

@rmehta 
@revant 
@RitvikSardana @nextchamp-saqib 

I need this because i was preventing to edit of non assigned users in a lead only assigned users can edit the lead

Shindhu-Ramaswamy:
Closes #1703

When a save fails (e.g. a server script calls frappe.throw), the error toast was shown multiple times on Lead and Deal pages. useDocument already shows the error, and updateField in the page components was showing it again. Removed the duplicate toast from the page components so only one error is shown.

### DIFF
--- a/frontend/src/data/document.js
+++ b/frontend/src/data/document.js
@@ -113,7 +113,7 @@ export function useDocument(doctype, docname, resourceOverrides = {}) {
                 toast.error(msg)
               })
 
-              if (err.messages?.length === 0) {
+              if (!err.messages?.length) {
                 toast.error(__('An error occurred while updating the document'))
               }
 
--- a/frontend/src/pages/Deal.vue
+++ b/frontend/src/pages/Deal.vue
@@ -778,13 +778,12 @@ function updateField(name, value) {
 
   document.save.submit(null, {
     onSuccess: () => (reload.value = true),
-    onError: (err) => {
+    onError: () => {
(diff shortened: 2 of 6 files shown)


ps173:
LGTM

PR #2822, opened 2026-09-11
Fix : quick filter input reset

### LINKED ISSUE #2113: Auto-refresh removes the digit during filtering
https://github.com/user-attachments/assets/6d759083-0ccd-46a2-94d6-0cd9fed6379e

When filtering records by Mobile Number in Frappe CRM list view, typing into the filter input sometimes removes the entered digits while the list auto-refreshes.

This happens more frequently when the dataset is large (observed with 900+ records).

#### Issue Behavior

* User starts typing mobile number in filter field
* List refreshes automatically during typing
* One or more typed digits disappear/reset from the input
* Makes filtering by mobile number unreliable/frustrating

#### Setup

* crm: 1.69.1

Shindhu-Ramaswamy:
closes : #2113 
Fix

ViewControls.vue 

Added reloadViewAfterSave(), which counts self-triggered view reloads (pendingSelfViewReloads) and decrements on failure so the counter cannot go stale.
createOrUpdateStandardView now calls reloadViewAfterSave() instead of reloadView().
The getView watcher skips reloading when a pending self-reload is outstanding. Our own save only replays state already applied locally, so there is nothing to rebuild. Reloads caused by external view changes still go through as before.

### DIFF
--- a/frontend/src/components/QuickFilterField.vue
+++ b/frontend/src/components/QuickFilterField.vue
@@ -37,10 +37,12 @@
   />
   <FormControl
     v-else
-    v-model="filter.value"
+    v-model="draft"
     type="text"
     :placeholder="filter.label"
-    @input.stop="debouncedFn(filter, $event.target.value)"
+    @focus="focused = true"
+    @blur="focused = false"
+    @input.stop="onTextInput($event.target.value)"
   />
 </template>
 <script setup>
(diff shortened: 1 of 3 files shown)


Shindhu-Ramaswamy:
[Screencast from 2026-09-20 21-33-39.webm](https://github.com/user-attachments/assets/9398d780-c9f8-4973-b28d-4b13d1eeeb52)

PR #2831, opened 2026-09-15
fix: complete web link attachment upload and surface its errors

### LINKED ISSUE #757: Attachment via Link Prevents Activities from Loading in Frappe CRM
### Steps to Reproduce:
1. Navigate to any Deal page in Frappe CRM.
2. Go to the Attachment tab.
3. Click the Upload Attachment button.
4. Select the Link option and input a valid link.
5. Click to attach button

### Issue Observed:
- After attaching a file via link, the Activities section does not load on the page.
- There is no update to any relevant state variable after successful attachment.
- Upload via link does not properly complete until the page is reloaded.
- The loading state remains true indefinitely after the upload, even though the attachment is successful.

### Suggested Fix:
**File** Path :- **_frontend/src/components/FilesUploader/FilesUploader.vue_**
**FIle Code Changes**
uploader.value
    .upload(file, args || {})
    .then(() => {
      if (i === files.value.length - 1 **_|| filesUploaderArea.value.showWebLink_**) {  _**// update this line 218**_ 
        files.value = []
        show.value = false
        fileUploadStarted.value = false
        emit('after')
      }
    })
    .catch((error) => {
      file.uploading = false
      let errorMessage = 'Error Uploading File'
      if (error?._server_messages) {
        errorMessage = JSON.parse(JSON.parse(error._server_messages)[0]).message
      } else if (error?.exc) {
        errorMessage = JSON.parse(error.exc)[0].split('\n').slice(-2, -1)[0]
      } else if (typeof error === 'string') {
        errorMessage = error
      }
      file.errorMessage = errorMessage
    })

Shindhu-Ramaswamy:
Fix

- Treat a web-link upload as the last upload (file.fileUrl || i === files.value.length - 1) so the success path resets state and emits `after`.
- Keep the dialog open until the upload finishes instead of closing it before the request.
- On web-link failure, reset the loading state and show the server error in a toast, since there is no file entry in the list to display it on.

Closes #757


Shindhu-Ramaswamy:
[Screencast from 2026-09-20 21-09-14.webm](https://github.com/user-attachments/assets/6bcfefee-6c44-4e49-8f36-b77a89204e99)

Issue #2841, opened 2026-09-16
BUG: Quick filter dropdown cuts off first option after selecting a bottom value


hiteshrabadia123:
**Describe the bug**
In the Organizations list view, the "No. of Employees" quick filter is a native `Select` field (frappe-ui `FormControl type="select"`). When the currently selected value is near the end of the option list, reopening the dropdown clips off the earlier options — they aren't rendered/visible at all.

**To Reproduce**
1. Go to CRM → Organizations list view.
2. Open the "No. of Employees" quick filter dropdown.
3. Select the last option in the list.
4. Re-open the dropdown and try to select the first option.

**Expected behavior**
All options should be visible/reachable in the reopened dropdown regardless of which option is currently selected.

**Screenshots**

<img width="1916" height="1007" alt="Image" src="https://github.com/user-attachments/assets/2566a9db-2fab-4f75-b1dc-88dbb19e5780" />

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

PR #2842, opened 2026-09-16
fix: prevent quick filter select dropdown from clipping options

Shindhu-Ramaswamy:
FIxes :#2841


Shindhu-Ramaswamy:
Screenshot from 2026-09-16 12-40-49

PR #2856, opened 2026-09-17
fix: reject CRM role changes for users with a Role Profile

### LINKED ISSUE #2321: Settings → Users: role change is silently ignored when the user has a Role Profile
## Description

In **CRM Settings → Users**, changing a user's role (or adding an existing user with a role) has **no effect** when that user has a **Role Profile** assigned. The UI shows success ("*{user} has been granted {role} access*"), but the role is never applied — the user's roles stay governed by their Role Profile.

CRM writes the role directly into the User's `roles` child table, but Frappe's `User.validate()` re-derives `roles` from the assigned Role Profile on every save, stripping anything not in the profile — so the change is reverted within the same save.

## Steps to reproduce

1. Assign a user to a Role Profile (User form → **Role Profiles**).
2. **CRM Settings → Users** → change that user's role via the dropdown.
3. See the success toast, then reload — the role was not applied.

## Expected

Either apply the change, or reject it with a clear message (as the *remove user* flow already does). Don't report success while doing nothing.

## Root cause

`crm/api/user.py` → `update_user_role()` appends the role and saves, with **no Role Profile check**:

```python
user_doc.append_roles("Sales Manager", "Sales User")
user_doc.save(ignore_permissions=True)   # triggers User.validate()
```

Frappe `frappe/core/doctype/user/user.py` → `populate_role_profile_roles()` (run on every save) strips it back out:

```python
self.roles = [r for r in self.roles if r.role in new_roles]   # removes CRM-appended roles
```

The frontend (`Users.vue` → `updateRole()`) shows `toast.success(...)` unconditionally, since the endpoint returns nothing on failure.

Inconsistency: `remove_crm_roles_from_user()` already guards this and throws ("*User cannot be removed as it has a Role Profile assigned*"); the add/change path has no equivalent guard.

## Affected code

- `crm/api/user.py` — `update_user_role()` (no guard), `add_existing_users()` (delegates to it)
- `crm/fcrm/doctype/crm_invitation/crm_invitation.py` — `accept()` uses the same `append_roles()` + `save()` pattern
- `frontend/src/components/Settings/Users.vue` — `updateRole()` reports success unconditionally

## Environment

- Frappe CRM `1.73.2`
- Frappe Framework `16.22.0`

Shindhu-Ramaswamy:
Closes #2321

Fix

crm/api/user.py: new validate_no_role_profile(user_doc) that throws when role_profiles (or the legacy role_profile_name) is set. Called at the start of update_user_role() — which also covers add_existing_users() — and reused by remove_crm_roles_from_user() in place of its inline check, so all three paths fail the same way.
crm_invitation.py: same check in before_insert (admin sees the error at invite time) and in accept() for existing users (profile assigned after the invite was sent).
No frontend change: Users.vue and AddExistingUserModal.vue already show error.messages[0] on failure, so the false success toast goes away on its own.
Error shown:
Roles of user [x@example.com](mailto:x@example.com) are managed by a Role Profile. Update the Role Profile instead.

Reproduce (before)

Role Profile → New: "Sales Only", roles = Sales User only.
User → any non-standard user → Role Profiles table → add "Sales Only" → save.
CRM → Settings → Users → change that user to Manager.
Success toast appears; reload → still Sales User.


Shindhu-Ramaswamy:
before :
[Screencast from 2026-09-17 12-17-39.webm](https://github.com/user-attachments/assets/55caaba2-75f0-492b-8ba2-8dda43d5a8db)

After : 
Screenshot from 2026-09-17 12-32-26

PR #2858, opened 2026-09-17
fix: drop stale fieldnames from saved layouts instead of passing strings to the UI

### LINKED ISSUE #2320: [Bug] "Tag" field type renders as Enter {0} and breaks in CRM Layout
**Describe the bug**
When attempting to add a standard Tag field to a Contact record using the modern CRM app's layout editor, the UI component fails to render correctly. It displays a broken string interpolation placeholder Enter {0} instead of the expected placeholder text, and the field entirely fails to trigger Frappe's native tag auto-complete or list lookup. Additionally, the field sometimes becomes unmanageable or disappears from the layout editor after breaking.

**To Reproduce**
	1.	Open the modern CRM app and navigate to a Contact record.
	2.	Click the Edit/Layout icon for the Details section.
	3.	Attempt to add or utilize a Tag field in the layout.
	4.	View the form UI.

**Expected behavior**
The field should display normal placeholder text and seamlessly hook into the core Frappe _user_tags system, offering the standard tag auto-complete dropdown when typing.
Screenshots

Image

**Environment:**
•	Self-hosted ERPNext / Frappe
•	App: Frappe CRM (latest)

Shindhu-Ramaswamy:
closes: #2320

Reproduce:
Customize Form → Contact → add a Data field (e.g. label Tag → custom_tag) → Update.
CRM → open a contact → Details → edit layout → add Tag → Save. Renders fine.
Delete the custom field (/app/custom-field/Contact-custom_tag).
Reload the contact: side panel is gone, console shows the TypeError. On a Lead the same steps give the Enter {0} slot.

Fix
Backend (crm_fields_layout.py): both resolvers now rebuild each column's field list from only the fieldnames that exist in meta; unresolved names are skipped. Also replaces the in-place list.index() mutation with a plain resolved_fields list. Permission handling, placeholders and the Required Fields logic are unchanged.
Frontend (Column.vue): filters out entries that aren't field objects before rendering, so a bad entry can never reach Field.vue from any path.
Tests (test_crm_fields_layout.py): two regression tests (Side Panel and Quick Entry) with a stale fieldname in the stored layout. They fail on develop (TypeError: string indices must be integers) and pass with this change.


Shindhu-Ramaswamy:
[Screencast from 2026-09-17 15-52-07.webm](https://github.com/user-attachments/assets/63cf0965-57e1-4298-acb3-e3bc74c3d985)

Issue #2857, opened 2026-09-17
Reply prefills From with the correspondent's address, so replies are sent spoofing the recipient


shahzeelahmed:
**Describe the bug**

Clicking **Reply** / **Reply All** on an email in the Activity timeline prefills the composer's `From` with the address of the email being replied to. That value is passed straight through as `sender` when sending, so outgoing replies leave with the *correspondent's* address in the `From` header, paired with the logged-in agent's display name:

```
From: Agent Full Name <lead@their-company.com>
```

When replying to a lead, the `From` becomes the lead's **own** address — the message appears to come from the person receiving it.

**To Reproduce**

1. Open a Deal or Lead that has an inbound email in the Activity timeline
2. Click **Reply**
3. Note the FROM selector now shows the correspondent's address rather than your own
4. Send the reply
5. Inspect the resulting Email Queue row (`sender`, and the `From:` header in `message`)

**Expected behavior**

The reply should be sent from the agent's own address (or an address they're actually permitted to send as), with the correspondent only in `To`.

**Cause**

`frontend/src/components/Activities/EmailArea.vue#L96-L97`:

```js
editor.fromEmail = email.sender   // <-- correspondent's address
editor.toEmails = [email.sender]  // correct
```

`toEmails` is right; `fromEmail` looks like an inadvertent duplicate of the line below it.

Then `frontend/src/components/CommunicationArea.vue#L207`:

```js
let fromEmail = newEmailEditor.value.fromEmail || getUser().email
```

Because `fromEmail` is now truthy, the `getUser().email` fallback never applies. Line 226 passes it as `sender`, while `sender_full_name` stays the logged-in user — which is why the name and address belong to different people.

The injected value isn't even guaranteed to be a valid option. The FROM dropdown is built from `user.doc.user_emails` (the `from` computed in `EmailEditor.vue`), and the watcher that validates `fromEmail`:

```js
watch(from, (fromOptions) => {
  if (!fromOptions.find((f) => f.value === fromEmail.value)) {
    fromEmail.value = fromOptions.length ? fromOptions[0].value : ''
  }
}, { immediate: true })
```

only fires when `from` changes, not when `fromEmail` is assigned. So an out-of-list address set by `reply()` persists unvalidated all the way to send.

**Impact**

This is primarily a deliverability bug, and it's close to invisible in the UI.

[...]

PR #2863, opened 2026-09-17
fix: prevent replies from spoofing the correspondent's email as sender

Lokesh-287:
## Issue
Reply prefills From with the correspondent's address, causing replies to be sent from an unpermitted/spoofed address

**Ref:** #2857

## Description
Clicking Reply on an inbound email set the composer's From to the correspondent's own address instead of the agent's mailbox. That value was passed straight through as `sender` on send, so outgoing replies left with the correspondent's address in the `From` header — in the reply case, From ends up equal to To, which mail providers treat as spoofing and drop silently, while Email Queue still shows "Sent".

## Fix
Removed the stray `editor.fromEmail = email.sender` line in `EmailArea.vue`'s `reply()`, and added a guard in `CommunicationArea.vue`'s `sendMail()` that only trusts `fromEmail` when it matches one of the agent's permitted mailboxes, falling back to their login email otherwise.

## Before
[Screencast from 09-18-2026 12:05:53 AM.webm](https://github.com/user-attachments/assets/be37a031-dc3f-42bc-9c6f-216fd242abdb)

## After
[Screencast from 09-18-2026 12:07:41 AM.webm](https://github.com/user-attachments/assets/98639cf2-3707-4fd4-b70c-4cdc17427796)


shahzeelahmed:
fixed with #2897

PR #2867, opened 2026-09-18
fix: show error toast when email template fails to render

### LINKED ISSUE #2621: Validation for email template parameters
**Describe the bug**
Add validation before saving email template parameters. Sometimes user can add parameters that are not really there and selection crashes. Later should also be gracefully handled

**To Reproduce**
1. Create a new template with a variable like `doc.first_name` in the body
2. Save and enable this template
3. Try to apply this template to lead/deal email reply 
4. Nothing happens + silent failure in logs

**Logs**
```
Uncaught (in promise) Error: frappe.email.doctype.email_template.email_template.get_email_template UndefinedError
    call call.ts:100
    applyEmailTemplate EmailEditor.vue:299
    callWithErrorHandling runtime-core.esm-bundler.js:199
    callWithAsyncErrorHandling runtime-core.esm-bundler.js:206
    emit runtime-core.esm-bundler.js:4443
    createSetupContext/get emit/< runtime-core.esm-bundler.js:8320
    onClick EmailTemplateSelectorModal.vue:29
    callWithErrorHandling runtime-core.esm-bundler.js:199
    callWithAsyncErrorHandling runtime-core.esm-bundler.js:206
    invoker runtime-dom.esm-bundler.js:730
call.ts:100:13
```

Shindhu-Ramaswamy:
Closes #2621

Problem

You could save an email template with a variable that doesn't exist, like lead.first_name or doc.emial. Saving worked, but when you tried to apply the template in the email editor, nothing happened. The error only showed up in the browser console.

What is already fixed

The exact example from the issue (doc.first_name) was fixed in #2753, which passes the doc nested as doc so both first_name and doc.first_name work. But any other wrong variable, like lead.first_name or a typo such as doc.emial, still fails silently. And the main ask in the issue, validating the variables before saving, was not done yet. This PR covers those remaining parts.

Fix

Both changes are in CustomEmailTemplate, which CRM already uses for Email Template. No new API was added.

Check variables when saving
On save, the subject and body are checked against the fields of the reference doctype. If a variable doesn't match any field, saving fails with a clear message, for example: CRM Lead has no field(s) named: doc.emial, lead. This is skipped if the template has no reference doctype.

Show a proper error when applying


Shindhu-Ramaswamy:
[Screencast from 2026-09-20 21-33-39.webm](https://github.com/user-attachments/assets/518c3baa-ad82-4b66-b684-c293c3b86434)
Screenshot from 2026-09-20 22-32-53

PR #2868, opened 2026-09-18
fix: keep select dropdown below trigger on email template pages

Shindhu-Ramaswamy:
Problem
In Settings > Email > Templates, the "Content Type" dropdown opens in a different place depending on the selected value.
When "Rich Text" is selected, the list opens over the field and looks like it opens downward.
After selecting "HTML", the list shifts up and looks like it opens upward, covering the field.
The same happens with the "For" dropdown once "Lead" is selected.

Fix
Pass side="bottom" to the "For" and "Content Type" selects on the New and Edit template pages. This switches the component to normal placement, so the list always opens below the field.


Shindhu-Ramaswamy:
[Screencast from 2026-09-20 21-04-29.webm](https://github.com/user-attachments/assets/210019eb-d3d3-4ea2-95ca-b7c62fe9217a)

PR #2876, opened 2026-09-21
fix: add tooltip to side panel edit layout button

Shindhu-Ramaswamy:
## Problem

The "edit fields layout" pencil button in the side panel of Lead, Deal, Contact and Organization pages has no tooltip on hover. Every other edit-layout pencil in the app (Data tab, quick-entry modals, grid) shows **Edit Fields Layout**; the side panel was the only one missing it.

## Fix

Pass `:tooltip="__('Edit Fields Layout')"` to the button in `SidePanelLayout.vue`. The string is already used and translated elsewhere, so no new locale entries are needed. Since the component is shared, this covers all four pages at once.


Shindhu-Ramaswamy:
Screenshot from 2026-09-21 15-25-33

PR #2879, opened 2026-09-22
fix: use NestedSet for CRM Territory so lft/rgt and tree filters work

Shindhu-Ramaswamy:


crm/patches/v1_0/rebuild_crm_territory_tree.py
```diff
@@ -0,0 +1,83 @@
...
+
+	if parents_with_children:
+		frappe.qb.update(territory).set(territory.is_group, 1).where(
+			territory.name.isin(parents_with_children) & (territory.is_group == 0)
+		).run()
+
+	rebuild_tree("CRM Territory")
+
+
+def break_parent_cycles():
```
shahzeelahmed: we can maybe remove this function and tests for it

PR #2884, opened 2026-09-22
fix: add error handling and Lost Reason prompt to Kanban drag-and-drop

Lokesh-287:
## Description
Dragging a Kanban card moves it visually right away, before the change is saved to the server. The save call had no error handling, so if the server rejected it — most commonly when dragging a Lead/Deal to a "Lost" status, which requires a reason the drag never provides — nothing told the user and the card just stayed put looking successful. Only a page refresh revealed it snapping back, since nothing was actually saved.

## Fix
The Kanban move call now properly handles both success and failure instead of firing and forgetting. If the server rejects a move for any reason, the card is reverted to its exact original column and position, and an error toast tells the user the move didn't go through — so failures are visible instead of silent.

For the one failure we can predict ahead of time — dragging a Lead/Deal into a "Lost" status, which always requires a reason — a dialog now asks for that reason before the save happens at all, instead of letting the save fail and cleaning up afterward. Confirming the dialog saves the status change and the reason together in one go; cancelling it puts the card back where it was, with a message explaining why.

Along the way, a related gap was also closed: when a card was dragged into a different column, its new position within that column was never being saved — only reordering cards *within* the same column was. Both cases are now handled the same way.

## Before

[Screencast from 09-22-2026 06:31:44 PM.webm](https://github.com/user-attachments/assets/c47dd0aa-4126-4ebb-a831-6b784793681c)

## After

[Screencast from 09-22-2026 06:35:37 PM.webm](https://github.com/user-attachments/assets/e7454475-45c0-4243-b6e9-243e2d33cda3)


frontend/src/components/Modals/KanbanLostReasonModal.vue
shahzeelahmed: instead of creating this new component, we can modify `LostReasonModal.vue` and reuse it here. You might just need to make `document` property optional. might also need to emit `save` and `cancel`. Then we wont need to add an identical component.

